Issue #3, opened 2016-05-26
Update to RC2


ciwchris:
Will this be updated to RC2 or is there another preferred way of reading settings from a configuration file? Currently I'm receiving the error:

> Reference to type 'ILibraryManager' claims it is defined in 'Microsoft.Extensions.PlatformAbstractions', but it could not be found

Because this is using "Microsoft.Extensions.PlatformAbstractions": "1.0.0-rc1-final"

nblumhardt:
Hi Chris - yes, definitely. There's a bit of a backlog of things to update for RC2 right now, trying to get to this one ASAP :-)

Cheers!

ciwchris:
Thanks, I thought that was probably the case. When RC2 released it created a lot of work. I see you've already been hard at work updating other libraries.

nblumhardt:
Almost there with this one; just need to figure out how to replace the use of  `ILibraryManager` and `PlatformServices.Default.LibraryManager` with `DependencyContext` in https://github.com/serilog/serilog-settings-configuration/blob/master/src/Serilog.Settings.Configuration/ConfigurationLoggerConfigurationExtensions.cs#L58 ...

nblumhardt:
@ciwchris PR linked above; it'd be great to have another set of eyes on it, if you have a chance to look it over. Shouldn't be far away now.

khellang:
Fixed with #4 :smile:

ciwchris:
Thanks @nblumhardt for such a quick turn around! Much appreciated!

PR #4, opened 2016-05-27
RC2 Update

nblumhardt:
A few changes here and there. `DependencyManager` seems to do the job of `ILibraryManager` well enough, but doesn't list assemblies that aren't directly referenced by the code, so loading based on the package name is necessary.

.NET 4.x now seems to need a build flag to preserve this info, so that's mentioned in the README.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Auto detect text files and perform LF normalization
+
+* text=auto
--- a/Build.ps1
+++ b/Build.ps1
@@ -1,103 +1,21 @@
-function Install-Dnvm
-{
-    & where.exe dnvm 2>&1 | Out-Null
-    if(($LASTEXITCODE -ne 0) -Or ((Test-Path Env:\APPVEYOR) -eq $true))
-    {
-        Write-Host "DNVM not found"
-        &{$Branch='dev';iex ((New-Object net.webclient).DownloadString('https://raw.githubusercontent.com/aspnet/Home/dev/dnvminstall.ps1'))}
-
-        # Normally this happens automatically during install but AppVeyor has
-        # an issue where you may need to manually re-run setup from within this process.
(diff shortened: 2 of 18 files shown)


nblumhardt:
Fixes #3.

nblumhardt:
cc @khellang, who keeps me honest when it comes to .NET Core best practices ;-)

ciwchris:
I pulled down the code, generated a NuGet package and installed it. Worked great! The only issue I ran into is when the tests run I receive the error:

> System.DllNotFoundException: Unable to load DLL 'Microsoft.DiaSymReader.Native.x86.dll': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
SUMMARY: Total: 2 targets, Passed: 1, Failed: 1.

I probably can look at it in more detail tomorrow

nblumhardt:
@ciwchris I think that issues is `dotnet`-version specific. Probably a non-issue for the time being, since there aren't any tests in this project (yet) :-)

khellang:
Yep. That issue is with the RC2 version of the dotnet xUnit runner. We encountered the same issues when updating Nancy. It happens when you target full framework.

To fix it, you have to add [this line](https://github.com/NancyFx/Nancy/blob/coreclr/test/Nancy.Embedded.Tests/project.json#L4) and move to [this version](https://github.com/NancyFx/Nancy/blob/coreclr/test/Nancy.Embedded.Tests/project.json#L3) of the runner. Make sure to also switch to the [preview feed](https://github.com/NancyFx/Nancy/blob/coreclr/NuGet.config#L8). See [the related issue](https://github.com/xunit/xunit/issues/843)

khellang:
LGTM :shipit: I think we can fix the xUnit runner in a separate PR :grin:

nblumhardt:
Thanks @khellang :+1:

PR #6, opened 2016-06-01
Read from config should not throw if MinimumLevel is not defined in settings file

skomis-mm:
IConfiguration.GetSection(..) doesn't return null if section is not defined. With MinimumLevel we should check it's Value property.


nblumhardt:
Thank you. I guess that makes a lot of the other null checks redundant, will look into removing those. Appreciate the PR!

Issue #20, opened 2016-08-22
Support MSBuild (non-project.json) projects


OzieGamma:
https://github.com/serilog/serilog-settings-configuration/blob/dev/src/Serilog.Settings.Configuration/ConfigurationLoggerConfigurationExtensions.cs

Source:
```csharp
context = DependencyContext.Default;
if (context == null)
  throw new InvalidOperationException("Add `\"preserveCompilationContext\": true` to `\"buildOptions\"` for support on .NET 4.x.");
```

When using this lib as part of a set-by-step upgrade from .NET 3.5 WinForms to .NET Core ASP.NET MVC I get the above exception to trigger. Any idea how to avoid it ?

nblumhardt:
Hi @OzieGamma - can you please include your `buildOptions` node from project.json? Thanks!

OzieGamma:
I am not building with .NET Core ...

nblumhardt:
Thanks for the update. It looks like the required `.deps.json` file is only generated by the `dotnet` tooling (project.json), not MSBuild, so for now it seems this library is only going to work with that project system.

I think with conditional compilation it should be possible to break the dependency this library has on `DependencyContext`, using `AppDomain` instead on the full framework. Marking this issue as up-for-grabs. Thanks!

PR #22, opened 2016-08-31
Support MSBuild (non-project.json) projects

skomis-mm:
fixes [this](https://github.com/serilog/serilog-settings-configuration/issues/20)


src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -154,27 +154,39 @@ Assembly[] LoadConfigurationAssemblies()
...
+            foreach (var assemblyName in GetSerilogConfigurationAssemblies())
             {
-                if (library.Name != null && library.Name.ToLowerInvariant().Contains("serilog"))
-                {
-                    var assumedName = new AssemblyName(library.Name);
-                    var assumed = Assembly.Load(assumedName);
-                    if (assumed != null && !assemblies.ContainsKey(assumed.FullName))
-                        assemblies.Add(assumed.FullName, assumed);
-
-                    foreach (var assemblyRef in library.Assemblies)
```
skomis-mm: Regarding this lines:
1) **Assemblies** collection is always empty (see [this](https://github.com/dotnet/cli/blob/rel/1.0.0/src/Microsoft.Extensions.DependencyModel/RuntimeLibrary.cs#L40-L45))
2) **RuntimeLibraries** has full graph in it (see updated Sample project: it references **Serilog.Sinks.RollingFile**, but configures **Serilog.Sinks.File** sink succesfully)

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -154,27 +154,39 @@ Assembly[] LoadConfigurationAssemblies()
...
+            var filter = new Func<string, bool>(name => name != null && name.ToLowerInvariant().Contains("serilog"));
+
+            if (_dependencyContext != null)
+            {
+                query = from lib in _dependencyContext.RuntimeLibraries where filter(lib.Name) select lib.Name;
+            }
+            else
+            {
+#if APPDOMAIN
+                query = from outputAssemblyPath in System.IO.Directory.GetFiles(AppDomain.CurrentDomain.BaseDirectory, "*.dll", System.IO.SearchOption.AllDirectories)
```
nblumhardt: `SearchOption.AllDirectories` might cause problems for apps that keep plug-ins in subfolders - often these contain distinct sets of assemblies that are loaded into different appdomains, meaning we could find multiple versions of DLLs of interest if this is specified.

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -154,27 +154,39 @@ Assembly[] LoadConfigurationAssemblies()
...
+                        let assemblyFileName = System.IO.Path.GetFileNameWithoutExtension(outputAssemblyPath)
+                        where filter(assemblyFileName)
+                        select Assembly.ReflectionOnlyLoadFrom(outputAssemblyPath).GetName().FullName;
+#endif
```
nblumhardt: Would an `#else` and throwing an exception be appropriate here? Not sure when it might occur, but a null `_dependencyContext` at this point might indicate an error.

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -154,27 +154,39 @@ Assembly[] LoadConfigurationAssemblies()
...
+            {
+                query = from lib in _dependencyContext.RuntimeLibraries where filter(lib.Name) select lib.Name;
+            }
+            else
+            {
+#if APPDOMAIN
+                query = from outputAssemblyPath in System.IO.Directory.GetFiles(AppDomain.CurrentDomain.BaseDirectory, "*.dll", System.IO.SearchOption.AllDirectories)
+                        let assemblyFileName = System.IO.Path.GetFileNameWithoutExtension(outputAssemblyPath)
+                        where filter(assemblyFileName)
+                        select Assembly.ReflectionOnlyLoadFrom(outputAssemblyPath).GetName().FullName;
```
nblumhardt: I may be missing the reason for a reflection-only load here; https://msdn.microsoft.com/en-us/library/system.reflection.assembly.reflectiononlyloadfrom.aspx suggests:

> You cannot execute code from an assembly that has been loaded into the reflection-only context. To execute the code, load the assembly with the LoadFile method.

nblumhardt:
Thanks for tackling this, looks like the right approach. Added a few comments inline.

Regards,
Nick

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
skomis-mm: Agree, removed. (experimented with custom probing paths, which would be weird with config assemblies in it)

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
skomis-mm: Just wanted to make GetSerilogConfigurationAssemblies() pure with best intent to take only metadata from the assembly and not load it with it references (the actual load in LoadConfigurationAssemblies()).
But found more appropriate AssemblyName.GetAssemblyName(path). Fixed.

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
skomis-mm: I think dependency context (and requirement to have preserveCompiltationContext=true) should be optional for several reasons:
1) it is already optional on public API (LoggerSettingsConfiguration.Configuration(..))
2) if some custom build tool doesn't generate dependency context, this library still has "Using" section, where we can put configuration assemblies without restriction to use this library (the source of this issue).
3) even if   dependencyContext is not null it doesn't help with non-conventional config assemblies that don't have "serilog" name in it. Again, we can put them in "Using" section.

nblumhardt: Makes sense, thanks for the explanation.

Issue #23, opened 2016-09-21
Environment-friendly syntax alternative


nblumhardt:
The current `"WriteTo"` syntax is intended to support multiple sinks of the same kind:

``` json
    "WriteTo": [
      { "Name": "Console" },
      { "Name": "File", "Args": { "path": "log.txt" } },
      { "Name": "File", "Args": { "path": "another.txt" } }
    ]
```

Unfortunately, in this syntax, representing (i.e. to override) settings using environment variables requires indexed access into the array, e.g.:

``` sh
set Serilog:WriteTo:1:Args:path=override.txt
# Or (bash):
export Serilog__WriteTo__1__Args__path=override.txt
```

Indexers like the `1` above are fragile and hard to guess.

An alternative syntax, that we might support side-by-side with the current one (effectively superseding it):

``` json
    "WriteTo": {
      "Console": {},
      "File": { "path": "log.txt" },
      "File2": { "path": "another.txt", "sink": "File" }
    }
```

It would be nice to be able to use something like `$sink` for the sink name in the full expansion, to avoid conflicting with sink parameters called `sink`, but unfortunately the valid character set across the target configuration sources is heavily limited.

This would in any case allow:

``` sh
set Serilog:WriteTo:File:path=override.txt
# Or:
export Serilog__WriteTo__File__path=override.txt
```

PR #28, opened 2016-10-14
Environment-friendly syntax alternative

skomis-mm:
- fixes [#23](https://github.com/serilog/serilog-settings-configuration/issues/23)
- adds xml/ini/json examples with multiple sinks of the same kind; updated Readme.md
- cosmetic changes/cleanup on project structure to align with other repos

PS regarding #23 `Microsoft.Extensions.Configuration` had already support for the explicitly named elements in the collections. Only for `json` its just a convention that every element has zero-based index for its name for simple arrays just to satisfy uniqueness. 
In contrast, `xml`/`ini` configurations must explicitly define for each element an unique name (see added examples).


nblumhardt:
Looks good - thanks for tackling this one! I need a few more days to dig into it, just letting you know it hasn't slipped off the radar.

trevorreeves:
+1 to support multiple sinks of the same kind.  The current version crashes in 'GetMethodCalls' as it tries to add duplicate keys to the result dictionary.

skomis-mm:
Closing this until [#37](https://github.com/serilog/serilog-settings-configuration/issues/37) and accompanying refactoring will be made on the codebase.

PR #44, opened 2017-04-05
Fix #957 [ReadFrom.Configuration(...) throws an exception on AWS..]

skomis-mm:
Contains 2 commits:
- vs2017 tooling update with some cleanup. Will reproduce issue on full .net in the test project.
- fixes [#957](https://github.com/serilog/serilog/issues/957) (when `Assembly.GetEntryAssembly()` is null):
  - workaround for `netcoreapp` - use `Using` section.
  - workaround for full .net - not required (will use conventions to lookup for configuration assemblies) or use `Using` section.


Build.ps1
```diff
@@ -20,7 +20,7 @@ foreach ($src in ls src/*) {
 
 	echo "build: Packaging project in $src"
 
-    & dotnet pack -c Release -o ..\..\artifacts --version-suffix=$suffix
+    & dotnet pack -c Release -o ..\..\.\artifacts --version-suffix=$suffix --include-source
```
nblumhardt: Is the `.\` redundant in this path?

skomis-mm: indeed :)

Issue #37, opened 2017-02-06
Configure sub-loggers through JSON appsettings


arielcloud:
something like:
```
"WriteTo": [
      {
        "Name": "Email",
        "Args": {
          ...
        },
        ***"Filter"***:  #....
      },
]
```
10x!

nblumhardt:
Regular Serilog configurations use `WriteTo.Logger()` for this:

```csharp
    .WriteTo.Logger(lc =>
        .Filter.ByIncludingOnly(...)
        .WriteTo.Email(...))
```

There's no equivalent syntax yet for configuration, mostly because we only just got a [configuration-compatible filtering provider](https://nblumhardt.com/2017/01/serilog-filtering-dsl/).

An extension to the JSON configuration syntax to do this would be great.

PR #46, opened 2017-04-05
Configure sub-loggers through JSON appsettings

skomis-mm:
- fixes [#37](https://github.com/serilog/serilog-settings-configuration/issues/37)
- adds `Serilog.Sinks.Async` support
- adds `LoggingLevelSwitch` parameter type support

Basically implementation based on converters around `Action<TSerilogConfiguration>` parameter types from configuration extensions methods.

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationSectionArgumentValue.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationSectionArgumentValue.cs
@@ -33,7 +33,7 @@ namespace Serilog.Settings.Configuration
                 return new Action<LoggerConfiguration>(_configReader.Configure);
             }
 
-            throw new Exception($"Handling {configurationType} is not implemented.");
+            throw new ArgumentException($"Handling {configurationType} is not implemented.");
         }
     }
 }
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationSectionArgumentValue.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationSectionArgumentValue.cs
@@ -1,0 +1,39 @@
+﻿using Serilog.Configuration;
+using System;
+using System.Reflection;
+
+namespace Serilog.Settings.Configuration
+{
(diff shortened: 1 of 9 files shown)


src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationSectionArgumentValue.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                return new Action<LoggerSinkConfiguration>(_configReader.ApplySinks);
+            }
+
+            if (configurationType == typeof(LoggerConfiguration))
+            {
+                return new Action<LoggerConfiguration>(_configReader.Configure);
+            }
+
+            throw new Exception($"Handling {configurationType} is not implemented.");
```
nblumhardt: Since this is wholly dependent on the value of `toType`, would `ArgumentException` be a reasonable choice here?

nblumhardt:
> adds LoggingLevelSwitch parameter type support

How does this one work in practice - is it possible to share a single level switch between the logger and one of the sinks (a-la Seq Sink's `controlLevelSwitch`)?

skomis-mm:
@nblumhardt , regarding `LoggingLevelSwitch`, I was thinking about defining switches and sharing them in the configuration. Not sure about it design so I did not included implementation for this.  

In this `pr` I just added converter from `string` to `LoggingLevelSwitch` for the extension method parameters types. For now it is just for `.Logger(..)` surrogate method that I added:
`internal static LoggerConfiguration Logger(LoggerSinkConfiguration loggerSinkConfiguration, Action<LoggerConfiguration> configureLogger, LoggingLevelSwitch restrictedToMinimumLevel = null)` so you can change level at runtime.

skomis-mm:
I've tested this for a while so I believe its ready to be in the wild. :bowtie:

nblumhardt:
Ah, I see now where `LoggingLevelSwitch` applies - thanks 👍

MelnikovIG:
@nblumhardt Add example of using this to documentation please.
https://github.com/serilog/serilog-settings-configuration

AraHaan:
Can ``Matching.FromSource`` be used in said configuration file when they want to filter with ``ByIncludingOnly``?

PR #48, opened 2017-04-11
Fix #47, #23 [multiple sinks of the same kind; environment-friendly syntax)

### LINKED ISSUE #47: Using multiple sinks of the same type from configuration
I would like to use multiple sinks of the same type (Async). I do want those sinks to be async independently of each other (without one being able to block the others). This works without using Configuration:
```C#
Log.Logger = new LoggerConfiguration()
    .ReadFrom.Configuration(Configuration)
    .WriteTo.Async(a => a.LiterateConsole())
    .WriteTo.Async(a => a.ElasticSearch())
    .WriteTo.Async(a => a.RollingFileAlternate())
    .CreateLogger();
```
Now, it is not possible to add them via Configuration as the Key "Async" can only be used once. The Async Sink provides a way of adding different sinks to it but this will make those sink not async from each other: 

```json
"WriteTo": [
      {
        "Name": "Async",
        "Args": {
          "configure": [
            {
              "Name": "LiterateConsole"
            },
            {
              "Name": "Elasticsearch",
              "Args": {
              }
            },
            {
              "Name": "RollingFileAlternate",
              "Args": {
              }
            }
          ]
        }
      }
    ]
```
Is this a question of Configuration or should Async Sink be able to set up multiple independent sinks?

skomis-mm:
- fixes [#47](https://github.com/serilog/serilog-settings-configuration/issues/47)
- fixes [#23](https://github.com/serilog/serilog-settings-configuration/issues/23)

#23 is a side effect of fixing #47 (as already pointed out in previous [pr](https://github.com/serilog/serilog-settings-configuration/pull/28) that syntax was already supported by the `Microsoft.Extensions.Configuration`. Fixing #47 just enables it.)

### DIFF
--- a/sample/Sample/appsettings.json
+++ b/sample/Sample/appsettings.json
@@ -8,38 +8,36 @@
         "MyApp.Something.Tricky": "Verbose"
       }
     },
-    "WriteTo": [
-      {
-        "Name": "Logger",
-        "Args": {
-          "configureLogger": {
-            "WriteTo": [
-              {
-                "Name": "LiterateConsole",
-                "Args": {
-                  "outputTemplate": "[{Timestamp:HH:mm:ss} {SourceContext} [{Level}] {Message}{NewLine}{Exception}"
-                }
-              }
-            ]
-          },
-          "restrictedToMinimumLevel": "Debug"
-        }
-      },
-      {
-        "Name": "Async",
-        "Args": {
-          "configure": [
+    "WriteTo:Sublogger": {
+      "Name": "Logger",
+      "Args": {
+        "configureLogger": {
+          "WriteTo": [
             {
-              "Name": "File",
+              "Name": "LiterateConsole",
               "Args": {
-                "path": "%TEMP%\\Logs\\serilog-configuration-sample.txt",
-                "outputTemplate": "{Timestamp:o} [{Level:u3}] ({Application}/{MachineName}/{ThreadId}) {Message}{NewLine}{Exception}"
+                "outputTemplate": "[{Timestamp:HH:mm:ss} {SourceContext} [{Level}] {Message}{NewLine}{Exception}"
               }
             }
           ]
-        }
-      }      
-    ],
+        },
+        "restrictedToMinimumLevel": "Debug"
+      }
+    },
+    "WriteTo:Async": {
+      "Name": "Async",
+      "Args": {
+        "configure": [
+          {
+            "Name": "File",
+            "Args": {
+              "path": "%TEMP%\\Logs\\serilog-configuration-sample.txt",
+              "outputTemplate": "{Timestamp:o} [{Level:u3}] ({Application}/{MachineName}/{ThreadId}) {Message}{NewLine}{Exception}"
+            }
+          }
+        ]
+      }
+    },
     "Enrich": ["FromLogContext", "WithMachineName", "WithThreadId"],
     "Properties": {
       "Application": "Sample"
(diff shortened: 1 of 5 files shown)


nblumhardt:
Great!

I'm having trouble following, you'll have to forgive an obvious question :-). In this example:

```json
{
    "WriteTo": [
      {
        "Name": "LiterateConsole",
        "Args": {
            "outputTemplate": "{Message}"
          }
      },
      "DiagnosticTrace"
    ],
    "WriteTo:File1": {
        "Name": "File",
        "Args": {
            "outputTemplate": "{Message}"
        }
    },
    "WriteTo:File2": {
        "Name": "File",
        "Args": {
            "outputTemplate": "{Message}"
        }
    }
}
```

Do the _m.e.Configuration_ library's constraints prevent us from just doing the following?

```json
{
    "WriteTo": [
      {
        "Name": "LiterateConsole",
        "Args": {
            "outputTemplate": "{Message}"
          }
      },
      "DiagnosticTrace",
      {
        "Name": "File",
        "Args": {
            "outputTemplate": "{Message}"
        }
      },
      {
        "Name": "File",
        "Args": {
            "outputTemplate": "{Message}"
        }
      }
   ]
}
```

skomis-mm:
@nblumhardt , no problem! I needed to clarify what means `Microsoft.Extensions.Configuration` supports "environment-friendly syntax". Your example will work and existing configurations will work as before.
   In `Microsoft.Extensions.Configuration` to support environment variables every path to configuration element must have unique name. In your example you use "array-like" syntax, which is just a convention and means that each element has section name based on its index:
```json
"WriteTo": [
{
   "Name": "LiterateConsole", 
   "Args": {
      "outputTemplate": "{Message}"
   }
}]
``` 
is equivalent to:
```json
"WriteTo:0": 
{
   "Name": "LiterateConsole", 
   "Args": {
      "outputTemplate": "{Message}"
   }
}
```

or

```json
"WriteTo": 
{
    "0": {
     "Name": "LiterateConsole", 
     "Args": {
        "outputTemplate": "{Message}"
      }
}
```

So your example could be (but not needed to) rewritten to this:
```json
"WriteTo:0": {
  "Name": "LiterateConsole",
  "Args": {
    "outputTemplate": "{Message}"
  }
},
"WriteTo:1": "DiagnosticTrace",
"WriteTo": {
  "2": {
    "Name": "File",
    "Args": {
      "outputTemplate": "{Message}"
    }
  }
},
"WriteTo:3": {
  "Name": "File",
  "Args": {
    "outputTemplate": "{Message}"
  }
}
```

That's it! And this convention about array indexes works only for the `json` package. In the other packages (`xml`, `ini`) you must always specify names for each element.
So what it means is that you can specify explicit names in `json` too:
```json
"WriteTo:Literate": {
  "Name": "LiterateConsole",
  "Args": {
    "outputTemplate": "{Message}"
  }
},
"WriteTo:Diag": "DiagnosticTrace",
"WriteTo": {
  "File1": {
    "Name": "File",
    "Args": {
      "outputTemplate": "{Message}"
    }
  }
},
"WriteTo:File2": {
  "Name": "File",
  "Args": {
    "outputTemplate": "{Message}"
  }
}
```
and `writeTo.GetChildren()` will return the same set of elements. But now reordering/removing elements will not change paths to them which is good when you override settings via environment variables.

skomis-mm:
Just to clarify more. These changes do not add anything new to support `environment-friendly` syntax. It was already there before, but not for the sinks of the same kind that were mentioned in the original issue #23 .

nblumhardt:
Ah, thanks - got it.

A little off-topic, but I keep wondering, semi-idly, how we would design the syntax if starting fresh today. The current syntax is fine, but does get quite verbose once `"Args"` is used. Probably not worth a change, since for minor aesthetic gain we'd break a lot of documentation (and SerilogAnalyzer), but still might be interesting to consider. I'll try to put my thoughts down in another ticket :-)

PR #56, opened 2017-05-17
2.4.0 Release

nblumhardt:
- #46 - configure sub-loggers through JSON settings
 - #48 - permit multiple sinks of the same kind


nblumhardt:
@skomis-mm I think this one's ready to go - any reason to hold off?

skomis-mm:
@nblumhardt , I don't think so. :shipit:

Issue #52, opened 2017-04-21
Misconfigured sinks fail silently


FantasticFiasco:
I would like to inquire the behavior where a misconfigured sink silently fails, in the aspect of not being created and configured in Serilog.

I'm on the verge of doing some work on two Serilog sink, work that would involve renaming arguments in the extension methods creating the sink. However I am afraid to do so because from the perspective of the consumers, even though I indicate my changes as breaking, they wouldn't notice at build, the application wouldn't throw an exception at runtime, the only behavioral change they would be able to notice would be the absence of log event from this particular sink.

What is the reasoning behind not throwing an exception when the JSON is misconfigured?

nblumhardt:
Hi @FantasticFiasco - the current thinking is that JSON configuration issues are in the "runtime" category - https://github.com/serilog/serilog/wiki/Reliability#configuration - not sure whether this is optimal, but had to pick one.

Would improving the `SelfLog` output generated from misconfigured sinks help here?

FantasticFiasco:
Would all JSON configuration errors have to fall into the same category? Wouldn't the full range of configuration errors in C# also be applicable to JSON?

From my perspective, JSON and C# configuration are just two sides of the same coin, and whether or not something is a _runtime configuration error_ or a _development-time invariant violation_ would be for the sink to decide.

Let me try to explain using the code example for the development-time invariant violation.

```
// Null is never acceptable as a filename, so
// Seriog will throw ArgumentNullException here.
Log.Logger = new LoggerConfiguration()
    .WriteTo.File(null)
    .CreateLogger();
```

Converting this to JSON and we would end up with something like this:

```
{
  "Serilog": {
    "WriteTo": [
      {
        "Name": "File",
        "Args": { }
      }
    ]
  }
}
```

Running these two configurations produce different behavior. The first will throw an exception, but the second won't, it will silently fail. I am guessing that the reason for this is that `Serilog.Settings.Configuration` can't find a suitable extension method since `path` is required, would that be a valid guess?

I think the point I am trying to make is that if `Serilog.Settings.Configuration` cannot delegate the responsibility to the sink, letting it determine whether the error is a runtime configuration error or a development-time invariant violation, then it should fall back to assume it to be a development-time invariant violation.

nblumhardt:
Thanks for the follow-up!

[...]

PR #67, opened 2017-09-06
Updated StringArgumentValue.ConvertTo to let the class not found exce…

formix:
…ption go

I lost a full afternoon looking for documentation or on stackoverflow to find out how to set an object instance to a method's "Args" in my appsettings.json file (same thing applies to the web.config I guess).

If I had recieved the correct exception, I would have known right away how to format the class name by looking at the stack trace and into the `Type.GetType` documentation.

I ran all unit tests and everything works fine with that change.

Hope it helps and you'll accept my pull request...

See my stackoverflow question [here](https://stackoverflow.com/questions/46082251/how-to-set-my-own-keygenerator-instance-in-appsettings-json) for some more background.

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/StringArgumentValue.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/StringArgumentValue.cs
@@ -56,7 +56,7 @@ namespace Serilog.Settings.Configuration
 
             if (toTypeInfo.IsInterface && !string.IsNullOrWhiteSpace(argumentValue))
             {
-                var type = Type.GetType(argumentValue.Trim(), throwOnError: false);
+                var type = Type.GetType(argumentValue.Trim());
                 if (type != null)
                 {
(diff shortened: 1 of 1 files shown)


nblumhardt:
Hi Jean Philippe, thanks for the PR!

I think you're right- looking at the other places in _Serilog.Settings.Configuration_ where we throw, it seems roughly in line with the rest of the library's behavior.

@serilog/reviewers-configuration any thoughts?

skomis-mm:
Looks great 👍 Good step to bring [strictValidation](https://github.com/serilog/serilog-settings-configuration/issues/52) eventially

formix:
W00t! My first accepted & merged pull request ever. I'll brag about that with my buddies. Just don't tell anyone that I changed just one line... shhhhhht.

Thanks and have a good one ;).

khellang:
Congrats, @formix! 🎉 🍰 🍾

PR #74, opened 2017-11-07
Added support for abstract classes

### LINKED ISSUE #71: Support a fully qualified type name for arguments whose type is abstract
Recently, the `LiterateConsole` sink was deprecated, pending replacement by the normal `Console` sink. To maintain backward functionality, the `Console` sink now supports a `theme` parameter (see [themes](https://github.com/serilog/serilog-sinks-console)). This parameter is an instance of `ConsoleTheme`, an abstract class. I expected to be able to use the fully qualified type name as is possible for `IFormatter`, but it gives an error of "string not castable to ConsoleTheme" because apparently, the string-to-type name resolution is only supported [for interfaces](https://github.com/serilog/serilog-settings-configuration/blob/dev/src/Serilog.Settings.Configuration/Settings/Configuration/StringArgumentValue.cs#L57). Is there a reason for this, or could that line be extended to include abstract (and perhaps even concrete) class types?

snoozbuster:
This should resolve #71.

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/StringArgumentValue.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/StringArgumentValue.cs
@@ -54,7 +54,7 @@ namespace Serilog.Settings.Configuration
             if (convertor != null)
                 return convertor(argumentValue);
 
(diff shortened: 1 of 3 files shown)


skomis-mm:
👍

Issue #73, opened 2017-11-07
Support for static member as the value of setting using `::`


tsimbalar:
This is the *Serilog.Settings.Configuration* transposition of serilog/serilog#1057 .

The idea is to support `"Serilog.Sinks.SystemConsole.Themes.AnsiConsoleTheme::Grayscale"` to access a static member for arguments of a complex type.

See details of issue in *core* *Serilog* repo to implement it in a consistent way.

This would help with serilog/serilog-sinks-console#16

tsimbalar:
Merged in #77

jp7677:
Thanks a lot! Could you consider to release a new version?

tsimbalar:
@jp7677 Hi, this is available in nuget package v2.5.0 which was (accidentally 😛 / by yours truly ) generated from the dev branch

PR #77, opened 2017-12-13
Support for static member as the value of setting using `::` #73

MrMon3y:
@tsimbalar 

Support "Serilog.Sinks.SystemConsole.Themes.AnsiConsoleTheme::Grayscale" to access a static member for arguments of a complex type.

Solves issue #73


tsimbalar:
This looks pretty good 👍  

Just a few remarks : 
- I believe all the classes under the `*.TestDummies.*` are not used anywhere, so they could be removed
- to "prove" the correct behavior, it would be nice to update the *Sample* project so that it uses the *Console* sink (instead of *Literate Console*)  and a custom theme with the new `::` syntax.

MrMon3y:
@tsimbalar 

**PR Update**
- Removed unused `TestDummies`
- Updated *Sample* project to use the *Console* sink and added a custom theme with new `::` syntax.

Should this also be reflected in the `readme.md` ?

sample/Sample/Sample.csproj
```diff
@@ -16,6 +16,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="1.1.1" />
     <PackageReference Include="Serilog.Sinks.Async" Version="1.0.1" />
+    <PackageReference Include="Serilog.Sinks.Console" Version="3.1.1" />
     <PackageReference Include="Serilog.Sinks.Literate" Version="2.0.0" />
```
tsimbalar: I think the package Serilog.Sinks.Literate is no longer required, because the sample app only use the Console Sink

test/Serilog.Settings.Configuration.Tests/Support/ClassHierarchy.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace Serilog.Settings.Configuration.Tests.Support
+{
+    public abstract class DummyAbstractClass
```
tsimbalar: I think this class (and its subclasses) are no longer needed. If this is the case, the file can probably be removed :)

tsimbalar:
Looks good ! 
I've added a few more comments, but I think this is close to being done 👍 

Regarding the edit to the `README.md`, I think it will sooner or later be included as part of serilog/serilog#1053 so that won't be necessary for now.

Maybe a comment in the `README.md` of [Serilog.Sinks.Console](https://github.com/serilog/serilog-sinks-console) would be nice though, but I'll create a separate issue for that one.

MrMon3y:
@tsimbalar 

**PR Update**
- Removed stale PackageReference
- Removed stale Class

tsimbalar:
Thanks Douglas, this looks good to me ! 👍
@nblumhardt @skomis-mm  : shall I merge it ?

nblumhardt:
Looks good to me! I see we are really overdue for unification of the various configuration systems - quite a bit of duplication here, but 👍 for now.

skomis-mm:
Looks good! 👍

PR #87, opened 2018-02-08
Support AuditTo in configuration

### LINKED ISSUE #80: Support for `AuditTo`
`LoggerConfiguration.AuditTo.*` was added to Serilog in v2.2. 

It has been available in *Serilog.Settings.AppSettings* for a while, but it does not seem to exist yet in *Serilog.Settings.Configuration¨*. 

The following tests fails : 

https://github.com/tsimbalar/serilog-settings-comparison/blob/ca5eee4cebcdf14b1f26b3c7e14ca63fd2a53412/test/Serilog.Settings.Comparison.Tests/2-Advanced%20scenarios/221_AuditToSink.cs#L17-L32

with config file : 
https://github.com/tsimbalar/serilog-settings-comparison/blob/ca5eee4cebcdf14b1f26b3c7e14ca63fd2a53412/test/Serilog.Settings.Comparison.Tests/2-Advanced%20scenarios/221-AuditToWithSimpleParams.json#L1-L15

Issue serilog/serilog#828 seems to state that JSON support is done, so I may be wrong :)

tsimbalar:
Fixes #80

- handles AuditTo
- adds more general tests with sample json config snippets
- update Serilog reference to v2.5.à

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -41,6 +41,7 @@ namespace Serilog.Settings.Configuration
             ApplyEnrichment(loggerConfiguration);
             ApplyFilters(loggerConfiguration);
             ApplySinks(loggerConfiguration);
+            ApplyAuditSinks(loggerConfiguration);
         }        
 
         void ApplyMinimumLevel(LoggerConfiguration loggerConfiguration)
@@ -95,6 +96,16 @@ namespace Serilog.Settings.Configuration
             {
                 var methodCalls = GetMethodCalls(writeToDirective);
                 CallConfigurationMethods(methodCalls, FindSinkConfigurationMethods(_configurationAssemblies), loggerConfiguration.WriteTo);
+            }
+        }
+
+        void ApplyAuditSinks(LoggerConfiguration loggerConfiguration)
+        {
+            var auditToDirective = _configuration.GetSection("AuditTo");
+            if (auditToDirective != null)
+            {
+                var methodCalls = GetMethodCalls(auditToDirective);
+                CallConfigurationMethods(methodCalls, FindAuditSinkConfigurationMethods(_configurationAssemblies), loggerConfiguration.AuditTo);
             }
         }
 
@@ -256,6 +267,13 @@ namespace Serilog.Settings.Configuration
             var found = FindConfigurationMethods(configurationAssemblies, typeof(LoggerSinkConfiguration));
             if (configurationAssemblies.Contains(typeof(LoggerSinkConfiguration).GetTypeInfo().Assembly))
                 found.Add(GetSurrogateConfigurationMethod<LoggerSinkConfiguration, Action<LoggerConfiguration>, LoggingLevelSwitch>((c, a, s) => Logger(c, a, s)));
+
+            return found;
+        }
+
+        internal static IList<MethodInfo> FindAuditSinkConfigurationMethods(IEnumerable<Assembly> configurationAssemblies)
+        {
+            var found = FindConfigurationMethods(configurationAssemblies, typeof(LoggerAuditSinkConfiguration));
 
             return found;
         }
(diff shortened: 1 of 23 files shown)


tsimbalar:
I don't know if it's worth incrementing the minor version -> v2.5.0 ... because of new feature, and also to align it with the version of Serilog that is referenced

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -260,6 +271,13 @@ internal static IList<MethodInfo> FindSinkConfigurationMethods(IEnumerable<Assem
             return found;
         }
 
+        internal static IList<MethodInfo> FindAuditSinkConfigurationMethods(IEnumerable<Assembly> configurationAssemblies)
+        {
+            var found = FindConfigurationMethods(configurationAssemblies, typeof(LoggerAuditSinkConfiguration));
```
tsimbalar: All the other methods somehow used `found.Add(GetSurrogateConfigurationMethod<T>` , but I was not sure whether that was needed for Audit ..

nblumhardt:
Thanks for the huge push on all of these things! I notice you're back from vacation :-)

I think updating the minor version is reasonable - though probably not a huge deal for downstream consumers.

Perhaps, since we're bumping the Serilog dependency version, we should update it to 2.6 so that the more recent package is brought in by default when installing this one? Thoughts?

tsimbalar:
Ok I'll increse the version number and reference v2.6.

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
skomis-mm: @tsimbalar `GetSurrogateConfigurationMethod<T>` is for methods mimicking static methods (extensions) and used just to get `MethodInfo` using expressions without reflection, but implementation is overcomplicated and I just realized you can get `MehtodInfo` right out of method name after casting to appropriate delegate type:
`(Func<LoggerFilterConfiguration, ILogEventFilter, LoggerConfiguration>loggerFilterConfiguration.With).MethodInfo`

skomis-mm:
Looks great! 👍 
(i'm a bit overwhelmed, but will keen to help as i can)

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
tsimbalar: oh I see. I guess it might not be necessary for `AuditTo`

tsimbalar:
btw, you may have noticed, but most of the test code is copied/pasted from the core *Serilog* repo :)

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
nblumhardt: I think it'd be needed for `AuditTo.Logger()` .. but this seems like a bit of a fringe case.

tsimbalar:
@nblumhardt @skomis-mm just a heads up, I think I may have messed up at some point, and added the xml element `<Version>` instead of updating the `<VersionPrefix>` in file `Serilog.Settings.Configuration.csproj` ... (becaue I updated the version number through VS's project properties UI)

I have added a commit to fix it afterwards (https://github.com/serilog/serilog-settings-configuration/commit/7d1f277de566fda38057bc79becf21a25e5d6de4) , but I think because of that change, a "non prerelease" nuget package has been published : https://www.nuget.org/packages/Serilog.Settings.Configuration/2.5.0 ... 

So now have a "non-prerelease" nuget package that has been produced from the `dev` branch instead of the `master` branch ... Not sure what's the fix for now ... Any idea ?

skomis-mm:
@tsimbalar I believe the only change should be is minor version bump (2.6) after [LoggingLevelSwitch](https://github.com/serilog/serilog-settings-configuration/pull/88) pr is merged since its adds new features. The published changes are minimal, tested, so there should be no problems.
We don't need to manage versioning since @nblumhardt handles this very well.

p.s. its nothing comparing how [CsvHelper](https://www.nuget.org/packages/CsvHelper/) was evolved in last few months :)

nblumhardt:
@tsimbalar I think the package out there as 2.5.0 is probably fine to leave, since it's been downloaded and used. Let's do as @skomis-mm suggests and bump the prefix to 2.6.0, which we can send through to `master` shortly after.

PR #95, opened 2018-04-05
Use `Console` instead of `LiterateConsole` in README

adamchester:
because [serilog-sinks-literate is being retired](https://github.com/serilog/serilog-sinks-literate#this-package-is-being-retired).


nblumhardt:
:+1:

Issue #83, opened 2018-01-27
Please update Microsoft.Extensions.DependencyModel to the latest version.


niemyjski:
This might break asp.net core 1.0 but if you are running on 2.0 you have major headaches and runtime errors as well as these warnings if you are using the `AspNetCore.All` meta package. And you are faced to add `<PackageReference Include="Microsoft.Extensions.DependencyModel" Version="2.0.3" />` of version >=2.0.3 

`warning MSB3277: Found conflicts between different versions of "Microsoft.Extensions.DependencyModel" that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed.`

nblumhardt:
:+1: this sounds like a fair change to make, at this point.

Help appreciated if anyone's able to work through it.

skomis-mm:
I didn't notice any compatibility at runtime and build issues with netcoreapp2.0 (tested `Microsoft.AspNetCore.All` 2.0.0 - 2.0.5 and sdks 2.0.0-2.1.4). Nuget resolves `Microsoft.Extensions.DependencyModel` to 2.0.0 since its closer to the app (it is a direct dependency of SDK) than 1.0.0.

nblumhardt:
Any chance you're using an old `dotnet` SDK version, @niemyjski ? I've found pretty substantial differences in behaviour from 2.0 up to 2.1.4.

zagros:
Any update on this?

nblumhardt:
@zagros any chance you can help pin this issue down to a particular configuration/SDK version/project setup? How are you reproducing? Thanks!

zagros:
This is happening on .NET Framework 4.6.2 for me the nuget package i'm using is 2.6.0

I enabled verbose logs and I then I saw this error

```
2>  Unified primary reference "Microsoft.Extensions.DependencyModel, Version=2.0.4.0, Culture=neutral, PublicKeyToken=adb9793829ddae60".
2>      Using this version instead of original version "1.0.0.0" in "C:\Work\_the_project\packages\Serilog.Settings.Configuration.2.6.0-dev-00081\lib\net451\Serilog.Settings.Configuration.dll" because AutoUnify is 'true'.
```

nblumhardt:
[...]

PR #96, opened 2018-04-10
project files updated

MV10:
Per #83, updated and tested references to `Microsoft.Extensions.DependencyModel` 
and `Microsoft.Extensions.Configuration.*`.

Confirmed both framework sample versions also run correctly. In all project files, old .NET Framework and .NET Standard refs were updated separately (conditional TFMs) to latest supporting versions. Also added a comment to the test project explaining how to run the multi-targeted tests from the command line (it was new to me, I suspect it is not commonly known).

### DIFF
--- a/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
+++ b/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
@@ -4,7 +4,7 @@
     <Description>Microsoft.Extensions.Configuration (appsettings.json) support for Serilog.</Description>
     <VersionPrefix>3.0.0</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>net451;netstandard1.6</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net451</TargetFrameworks>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AssemblyName>Serilog.Settings.Configuration</AssemblyName>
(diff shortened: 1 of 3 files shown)


MV10:
I've never used Travis, not sure why it blows up with a `netstandard1.6` error... there is nothing left in any part of the project which references it. Everything else works, hopefully it's an easy fix.

Also, didn't bump the version, not sure whether you'd consider this a minor or patch change.

> `/usr/share/dotnet/sdk/2.1.4/Sdks/Microsoft.NET.Sdk/build/Microsoft.PackageDependencyResolution.targets(167,5): error : Assets file '/home/travis/build/serilog/serilog-settings-configuration/src/Serilog.Settings.Configuration/obj/project.assets.json' doesn't have a target for '.NETStandard,Version=v1.6'. Ensure that restore has run and that you have included 'netstandard1.6' in the TargetFrameworks for your project. [/home/travis/build/serilog/serilog-settings-configuration/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj]`

adamchester:
Hey @MV10, you should try changing [buld.sh](https://github.com/serilog/serilog-settings-configuration/blob/dev/build.sh#L8) which is locked to `netstandard1.6`

MV10:
@adamchester Thanks! The file isn't in the SLN view and I didn't think to jump back to the file view. All good now, much appreciated.

src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
```diff
@@ -16,18 +16,22 @@
...
+  <ItemGroup Condition="'$(TargetFramework)' == 'net451'">
+    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="1.1.2" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="2.0.1" />
+  </ItemGroup>
+
   <PropertyGroup Condition=" '$(TargetFramework)' == 'net451' ">
     <DefineConstants>$(DefineConstants);APPDOMAIN</DefineConstants>
```
nblumhardt: Are the features behind `#if APPDOMAIN` now usable on .NET Standard 2.0?

nblumhardt:
Ah, great 👍 - that code enables using types without listing all of the assemblies in `serilog:using` keys on the full framework. It probably will kick in when targeting .NET Framework using the .NET Standard 2.0 targets, so perhaps we remove the `#if` and enable it on all platforms? I'm reasonably certain `AppDomain.BaseDirectory` is available on .NET Standard 2.

MV10:
@nblumhardt Good call, you're right. I'd moved `netstandard2_0` to be the first TFM which makes VS default to that, so as a first look Intellisense was satisfied without the constant. Builds ok, tests pass, Travis is happy.

Not sure why I thought `System.IO` was framework-level rather than standard-API. I suppose I haven't had much call for raw file I/O lately. Anyway, I think we're good to go: newer libs and less code, can't argue with that!

PR #100, opened 2018-04-19
Fixes #97, #83, #98

### LINKED ISSUE #83: Please update Microsoft.Extensions.DependencyModel to the latest version.

### LINKED ISSUE #97: Populate IConfiguration or IConfigurationSection parameter
## Proposal

The package will automatically populate any `IConfiguration` or `IConfigurationSection` parameter on the target method.

## The Issue

Any library being initialized by this package can only receive literals defined in the `WriteTo` entries. If the target supports more complex configuration than `WriteTo` provides (for example, currently [arrays are not supported](https://github.com/serilog/serilog-settings-configuration/blob/dev/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs#L372)), the target has no means to access _any_ configuration data when the target method is executed.

(This was not an issue under the .NET Framework in packages such as `Serilog.Settings.AppSettings` because the old `ConfigurationManager` approach is a static class, which means it is always available. `Microsoft.Extensions.Configuration` has no equivalent facility, it is the responsibility of the referencing library or application to provide access.)

In theory this issue could be avoided by simply not using this package at all, but I argue this limitation won't be obvious to Serilog users. For example, the MS SQL sink can add custom columns via a separate configuration section, so using this package is mutually exclusive with that feature. Each target package could explain any such limitations, but it seems easier in the long run to just provide a means to fix the problem in a way that is transparent to Serilog users.

If we make this package "smart" enough to populate `IConfiguration` or `IConfigurationSection` then, as dependent Serilog packages are updated to full `Microsoft.Extensions.Configuration` support, the problem basically fixes itself. (Probably the README for this package should provide advice to prefer `IConfiguration` over `IConfigurationSection` but that isolates RTFM caveats to the docs in this single repo.)

## Implementation

Currently, the log config extension accepts either `IConfiguration` or `IConfigurationSection`, but `IConfiguration` simply retrieves the Serilog section and calls the section-based extension method. The section is passed into `ConfigurationReader` where all of the main processing occurs when the main Serilog package calls the `ILoggerSettings.Configure` method.

I propose to create two ctor overloads which also accept `IConfiguration`. They will extract the Serilog section and pass it along to the existing ctors. All ctors will store a local reference to `IConfiguration` (if available) and `IConfigurationSection`.

Finally, `CallConfigurationMethods` would be modified to check each `MethodInfo` for an `IConfigurationSection` parameter and an `IConfiguration` parameter (when available), then add these to the `call` list before invoking the method.

A pair of related tests should also be added, and possibly the sample could be updated. 

"All ya gotta do is..."

And yes, I'm proposing to do the work myself, although any Reflection Ninjas in the house are more than welcome to handle it (I'm a little rusty on the specifics).

### LINKED ISSUE #98: Unable to specify a string array
Hi,

I'm using `Serilog.Sinks.AzureTableStorage` package and more specifically, the `WriteTo.AzureTableStorageWithProperties`, which accepts a `propertyColumns` argument, which is of type `string[]`.

No matter what I tried, I can't make it work so that I can pass an array of strings in my `appsettings.json` (or using the colon-separated syntax from my environment variables).

I tried doing something like this in `appsettings.json`:

```js
WriteTo: [
  // ...
  {
    "Name": "AzureTableStorageWithProperties",
    "Args": {
      // ...
      "propertyColumns": [ "Column1", "Column2" ]
    }
  }
]
```

But I get an error saying that `Argument value should be of type Action<>`.

I tried this: `"propertyColumns": "Column1, Column2"`, but this one also blows.

I also tried using environment variables, e.g.:

```
Serilog:WriteTo:2:Args:propertyColumns:0="Column1"
Serilog:WriteTo:2:Args:propertyColumns:1="Column2"
```

But, again no luck :(

So, is it possible somehow to specify a string array using this library? If so, what am I missing here?

I would really, really appreciate if you help me out here

MV10:
Fixes #97, adds support for `IConfiguration` parameters, `IConfigurationSection` parameters, config-binding of sections by parameter type, and fixes #83, updates MS DependencyModel and Configuration package versions. Probably also addresses #98 though I haven't tested it with arrays specifically. _Update: Confirmed, should fix #98._

- README updated
- `GetSection` never returns null; corrected null-ref tests to `GetChildren().Any()` tests
- ~~Added `>` "hint" character to identify nested configuration sections~~
- Added `ObjectArgumentValue` class to perform config binding and populate config section parameters
- Added two lines of code just before method invocation to populate an `IConfiguration` parameter

**Updated:**
- Moved nested configuration functionality into `ObjectArgumentValue`
- Removed `ConfigurationSectionArgumentValue`
- Added unit test for `IConfiguration` and `IConfigurationSection` arguments
- Added unit test for configuration object binding argument (for example, `List<>`)
- Added unit tests for string and int array arguments

SQL Sink PR depends on this:
https://github.com/serilog/serilog-sinks-mssqlserver/pull/120


MV10:
Oh, joy... I get to figure out those nasty unit tests after all... 👎

README.md
```diff
@@ -91,3 +91,41 @@ For example, to set the minimum log level using the _Windows_ command prompt:
 set Serilog:MinimumLevel=Debug
 dotnet run
 ```
+
+### Nested configuration sections
+
+Some Serilog packages require a reference to a logger configuration object. These must be "hinted" by appending a `>` character to the end of the argument name. The sample program in this project illustrates this with the following entry configuring the _Serilog.Sinks.Async_ package to wrap the _Serilog.Sinks.File_ package.
```
nblumhardt: This seems like a bit of a UX regression, and it's not immediately clear when this will or won't be a breaking change; can't we work out what kind of argument we're dealing with by reflecting over the parameters? I am afraid I'm missing something here, sorry - struggling to keep up :-)

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -19,25 +19,45 @@ namespace Serilog.Settings.Configuration
...
 
-        readonly IConfigurationSection _configuration;
+        static IConfiguration _configuration;
+
+        readonly IConfigurationSection _section;
         readonly DependencyContext _dependencyContext;
         readonly IReadOnlyCollection<Assembly> _configurationAssemblies;
 
-        public ConfigurationReader(IConfigurationSection configuration, DependencyContext dependencyContext)
+        #region Constructors
```
nblumhardt: Nit: the Serilog projects don't use `#region` at all :-)

README.md
MV10: I'll take another look, but other than the Async wrapper package, frankly I'm not clear on the use cases for nested config. (For example, the reason for the logger nested config in the sample code and tests makes no sense to me.) I don't really like the hack myself, but at the point the code is "deciding" how to store the config section (the parameter value, effectively), we don't yet know what the parameter type is. "How to store" means the various parameter-value classes (string, config, or my new object class). I suspect the config value class can go away and that can be added to the object value class.

README.md
```diff
@@ -91,3 +91,41 @@ For example, to set the minimum log level using the _Windows_ command prompt:
...
+
+If the parameter value is not a discrete value, the package will use the configuration binding system provided by _Microsoft.Extensions.Options.ConfigurationExtensions_ to attempt to populate the parameter. Almost anything that can be bound by `IConfiguration.Get<T>` should work with this package. An example of this is the optional `List<Column>` parameter used to configure the .NET Standard version of the _Serilog.Sinks.MSSqlServer_ package.
+
+### IConfigurationSection parameters
+
+Certain Serilog packages may require configuration information that can't be easily represented by discrete values or direct binding-friendly representations. An example might be lists of values to remove from a collection of default values. In this case the method can accept an entire `IConfigurationSection` as a call parameter and this package will recognize that and populate the parameter. In this way, Serilog packages can support arbitrarily complex configuration scenarios.
```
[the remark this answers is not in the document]
MV10: I think in the long run that one is going to be very useful to sink authors.

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
```diff
@@ -402,7 +402,7 @@ public void WriteToLoggerWithRestrictedToMinimumLevelIsSupported()
                 ""WriteTo"": [{
                     ""Name"": ""Logger"",
                     ""Args"": {
-                        ""configureLogger"" : {
+                        ""configureLogger>"" : {
```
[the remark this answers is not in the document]
MV10: That test already checks nested configs (there are two of them). Not clear what else you're asking for. However, it probably does need a test for `IConfigurationSection` and maybe another for config binding, such as `List<T>`... I'm still not clear how some of the test "dummy" sinks actually work though, seems overly complicated. I wish there were comments. But I'll have a look.

README.md
MV10: Fixed. Very glad you questioned this -- the "hint" hack is gone and the code is simplified! I took another look at what was going on in `ConfigurationSectionArgumentValue` and it turns out I could get rid of that and move the functionality into `ObjectArgumentValue`. The trick was deferring the new `ConfigurationReader` pass to read the nested config until we are actually calling the target method. Now the read is triggered by the `Action<>` parameter's type rather than "guessing" based on config structure.

This made me notice an edge-case risk of error based on how this package uses the configuration section `Value` property that I will open an issue about just to document it (nothing relating to my changes). It's extremely unlikely to occur in the wild but maybe somebody will feel like addressing it.

nblumhardt:
Awesome PR! Just going to check that the build goes through with my minor CSPROJ tweak, then will merge.

The PR comment "disappearing" trick is a bit of an annoying UI default in GitHub - once the code in question has changed, the comments on it are hidden by default.

Cheers!

PR #105, opened 2018-04-28
Multiple value exception, fixes #103

### LINKED ISSUE #103: Minor issue relating to ConfigurationSection.Value property
The way the package uses `ConfigurationSection.Value` has the potential to produce unexpected and incorrect behavior. I say "minor issue" because the configuration setup that would lead to this seems to be a very unlikely edge-case (or outright incorrect). I'm documenting it in case someone encounters it or feels like figuring out a fix. This will be more relevant once PR https://github.com/serilog/serilog-settings-configuration/pull/100 is merged.

In several places the code reads a configuration section, then tests the `Value` property for null. If it is not null, the code processes the property as a discrete string value. If the property is null, the code assumes an entire new configuration section has been retrieved (which may later be resolved as one of several complex types). The code always performs this test with an if/else structure.

The problem is that it's possible for _both_ scenarios to be true simultaneously: `Value` can have a discrete string value _and_ the section can have child entries. The MS Configuration package allows the application to overlay settings from multiple configuration sources. One source might assign a discrete value while a second source assigns a complex value.

I believe it's an edge-case because I can't think of a configuration scenario where that's likely to be needed.

In theory, something could take an `object` argument which could then accept a `string` or something like a `List<>`, but regardless, supplying two values for a single argument doesn't make any sense. There is no way to tell which was applied last, so we can't apply LIFO logic. Therefore I think the fix would be to throw an exception if both `Value` and `GetChildren().Any()` are provided for a single argument.

Offhand I don't remember what happens anywhere else this test is performed, so I'm unsure what may be needed to address those.

Up for grabs! 😀

MV10:
Throws `InvalidOperationException` if multiple configuration sources results in an ambiguous argument value. See #103 for details.

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -222,8 +222,14 @@ namespace Serilog.Settings.Configuration
             {
                 IConfigurationArgumentValue argumentValue;
 
+                // Reject configurations where an element has both scalar and complex
+                // values as a result of reading multiple configuration sources.
                 if (argumentSection.Value != null && argumentSection.GetChildren().Any())
-                    throw new InvalidOperationException($"Combined configuration sources must result in a discrete value (string, int, etc.) or complex value (section, list, etc.), not both. Argument: {argumentSection.Path}");
+                    throw new InvalidOperationException(
+                        $"The value for the argument {argumentSection.Path} is assigned different value " +
+                        "types in more than one configuration source. Ensure all configurations consistently " +
+                        "use either a scalar (int, string, boolean) or a complex (array, section, list, " +
+                        "POCO, etc.) type for this argument value.");
 
                 if (argumentSection.Value != null)
                 {
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -226,7 +226,7 @@ namespace Serilog.Settings.Configuration
                 // values as a result of reading multiple configuration sources.
                 if (argumentSection.Value != null && argumentSection.GetChildren().Any())
                     throw new InvalidOperationException(
-                        $"The value for the argument {argumentSection.Path} is assigned different value " +
+                        $"The value for the argument '{argumentSection.Path}' is assigned different value " +
                         "types in more than one configuration source. Ensure all configurations consistently " +
                         "use either a scalar (int, string, boolean) or a complex (array, section, list, " +
                         "POCO, etc.) type for this argument value.");
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -221,6 +221,10 @@ namespace Serilog.Settings.Configuration
             IConfigurationArgumentValue GetArgumentValue(IConfigurationSection argumentSection)
             {
                 IConfigurationArgumentValue argumentValue;
+
+                if(argumentSection.Value != null && argumentSection.GetChildren().Any())
+                    throw new InvalidOperationException($"Combined configuration sources must result in a discrete value (string, int, etc.) or complex value (section, list, etc.), not both. Argument: {argumentSection.Path}");
+
                 if (argumentSection.Value != null)
                 {
                     argumentValue = new StringArgumentValue(() => argumentSection.Value, argumentSection.GetReloadToken);
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -222,7 +222,7 @@ namespace Serilog.Settings.Configuration
             {
                 IConfigurationArgumentValue argumentValue;
 
-                if(argumentSection.Value != null && argumentSection.GetChildren().Any())
+                if (argumentSection.Value != null && argumentSection.GetChildren().Any())
                     throw new InvalidOperationException($"Combined configuration sources must result in a discrete value (string, int, etc.) or complex value (section, list, etc.), not both. Argument: {argumentSection.Path}");
 
                 if (argumentSection.Value != null)
--- a/test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
+++ b/test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
@@ -562,7 +562,7 @@ namespace Serilog.Settings.Configuration.Tests
 
         [Trait("Bugfix", "#103")]
         [Fact]
-        public void MultipleArgumentValuesThrowsInvalidOperationException()
+        public void InconsistentComplexVsScalarArgumentValuesThrowsIOE()
         {
             var jsonDiscreteValue = @"{
                 ""Serilog"": {            
--- a/test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
+++ b/test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
@@ -594,8 +594,8 @@ namespace Serilog.Settings.Configuration.Tests
             var ex = Assert.Throws<InvalidOperationException>(()
                 => ConfigFromJson(jsonDiscreteValue, jsonComplexValue));
 
-            Assert.Contains("Combined configuration sources", ex.Message);
-            Assert.Contains("pathFormat", ex.Message);
+            Assert.Contains("The value for the argument", ex.Message);
+            Assert.Contains("'Serilog:WriteTo:0:Args:pathFormat'", ex.Message);
         }
     }
 }
--- a/test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
+++ b/test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
@@ -12,9 +12,11 @@ namespace Serilog.Settings.Configuration.Tests
 {
     public class ConfigurationSettingsTests
     {
-        static LoggerConfiguration ConfigFromJson(string jsonString)
+        static LoggerConfiguration ConfigFromJson(string jsonString, string secondJsonSource = null)
         {
-            var config = new ConfigurationBuilder().AddJsonString(jsonString).Build();
+            var builder = new ConfigurationBuilder().AddJsonString(jsonString);
+            if(secondJsonSource != null) builder.AddJsonString(secondJsonSource);
+            var config = builder.Build();
             return new LoggerConfiguration()
                 .ReadFrom.Configuration(config);
         }
@@ -555,6 +557,43 @@ namespace Serilog.Settings.Configuration.Tests
             log.Write(Some.WarningEvent());
 
             Assert.Equal(1, DummyRollingFileSink.Emitted.Count);
+        }
+
+        [Trait("Bugfix", "#103")]
+        [Fact]
+        public void MultipleArgumentValuesThrowsInvalidOperationException()
+        {
+            var jsonDiscreteValue = @"{
+                ""Serilog"": {            
+                    ""Using"": [""TestDummies""],
+                    ""WriteTo"": [{
+                        ""Name"": ""DummyRollingFile"",
+                        ""Args"": {""pathFormat"" : ""C:\\""}
+                    }]        
+                }
+            }";
+
+            var jsonComplexValue = @"{
+                ""Serilog"": {            
+                    ""Using"": [""TestDummies""],
+                    ""WriteTo"": [{
+                        ""Name"": ""DummyRollingFile"",
+                        ""Args"": {""pathFormat"" : { ""foo"" : ""bar"" } }
+                    }]        
+                }
+            }";
+
+            // These will combine into a ConfigurationSection object that has both
+            // Value == "C:\" and GetChildren() == List<string>. No configuration
+            // extension matching this exists (in theory an "object" argument could
+            // accept either value). ConfigurationReader should throw as soon as
+            // the multiple values are recognized; it will never attempt to locate
+            // a matching argument.
+
+            var ex = Assert.Throws<InvalidOperationException>(() => ConfigFromJson(jsonDiscreteValue, jsonComplexValue));
+
+            Assert.Contains("Combined configuration sources", ex.Message);
+            Assert.Contains("pathFormat", ex.Message);
         }
     }
 }
--- a/test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
+++ b/test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
@@ -15,7 +15,8 @@ namespace Serilog.Settings.Configuration.Tests
         static LoggerConfiguration ConfigFromJson(string jsonString, string secondJsonSource = null)
         {
             var builder = new ConfigurationBuilder().AddJsonString(jsonString);
-            if(secondJsonSource != null) builder.AddJsonString(secondJsonSource);
+            if (secondJsonSource != null)
+                builder.AddJsonString(secondJsonSource);
             var config = builder.Build();
             return new LoggerConfiguration()
                 .ReadFrom.Configuration(config);
@@ -33,7 +34,7 @@ namespace Serilog.Settings.Configuration.Tests
                     }
                 }
             }";
-            
+
             var log = ConfigFromJson(json)
                 .WriteTo.Sink(new DelegatingSink(e => evt = e))
                 .CreateLogger();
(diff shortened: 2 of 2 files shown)


src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -221,6 +221,10 @@ void ApplyEnrichment(LoggerConfiguration loggerConfiguration, IReadOnlyDictionar
             IConfigurationArgumentValue GetArgumentValue(IConfigurationSection argumentSection)
             {
                 IConfigurationArgumentValue argumentValue;
+
+                if(argumentSection.Value != null && argumentSection.GetChildren().Any())
+                    throw new InvalidOperationException($"Combined configuration sources must result in a discrete value (string, int, etc.) or complex value (section, list, etc.), not both. Argument: {argumentSection.Path}");
```
bartelink: rest of code uses redundant braces ;)

bartelink: please use standard formatting (ctrl K D) can do it - e.g. `if (` should have a space

bartelink: Given this has the same precondition, can move inside the other if to L230

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
```diff
@@ -556,5 +558,42 @@ public void WriteToSubLoggerWithLevelSwitchIsSupported()
...
+            }";
+
+            // These will combine into a ConfigurationSection object that has both
+            // Value == "C:\" and GetChildren() == List<string>. No configuration
+            // extension matching this exists (in theory an "object" argument could
+            // accept either value). ConfigurationReader should throw as soon as
+            // the multiple values are recognized; it will never attempt to locate
+            // a matching argument.
+
+            var ex = Assert.Throws<InvalidOperationException>(() => ConfigFromJson(jsonDiscreteValue, jsonComplexValue));
```
bartelink: can you split the line please?

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
```diff
@@ -12,9 +12,11 @@ namespace Serilog.Settings.Configuration.Tests
 {
     public class ConfigurationSettingsTests
     {
-        static LoggerConfiguration ConfigFromJson(string jsonString)
+        static LoggerConfiguration ConfigFromJson(string jsonString, string secondJsonSource = null)
         {
-            var config = new ConfigurationBuilder().AddJsonString(jsonString).Build();
+            var builder = new ConfigurationBuilder().AddJsonString(jsonString);
+            if(secondJsonSource != null) builder.AddJsonString(secondJsonSource);
```
bartelink: please autoformat

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
bartelink: On reflection - the rest of this file is light on redundant braces; please disregard. would be nice if the exception message can be split across lines (example from further down:

```
  throw new InvalidOperationException(
                            "A zero-length or whitespace assembly name was supplied to a Serilog.Using configuration statement.");
```

MV10: Agreed about the if-space -- not my style but definitely seems to be the Serilog way.

On if/else I don't think braces are redundant. I eliminate on stand-alone one-liner ifs, but plenty of other if/else statements elsewhere in the code with braces that could be eliminated.

I think it's much clearer to perform the test separately (my guess is the compiler will optimize anyway, not that it would be measurable).

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
MV10: Doesn't match other `Assert.Throws` in the tests or many other lambdas throughout the project, but it doesn't bother me either way, I'll update.

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
bartelink: Fair point re it being subjective whether moving it inside there makes sense;  I'll leave it to your judgement.

MV10:
@bartelink Thanks for the review. Are Serilog code standards posted somewhere?

I just realized after I push my edits, my replies to your review comments will go away, so copying here:

- Agreed about space after `if` -- not my style but definitely the Serilog way.
- Disagree about `if/else` redundant braces. Agree about one-liner `if` but there are plenty of other `if/else` blocks in the code where one could theoretically remove the braces.
- Disagree about moving the exception test inside the following `if` -- I believe it is clearer to test for this separately, and I'm sure the compiler will optimize this anyway (not that the performance diff would be measurable).

- I'll split the `Assert.Throws` lambda (that's my preferred style, too), but it doesn't match other `Assert.Throws` or other lambdas in general elsewhere in the code.

- JSON utility function reformatted.

MV10:
@bartelink Maybe I can explain the issue more clearly.

_Microsoft.Extensions.Configuration_ allows an application to load configuration information from multiple sources. The most common example is when an application has an `appsettings.json` file and an `appsettings.development.json` file. The dev file is loaded in debug and typically contains a subset of settings which overrides the production settings (because it is loaded after the production version).

Consider these two simple bits of JSON (an extremely oversimplified example):

```json
{
    "setting" : "discrete"
}
```

versus

```json
{
    "setting" : [ "foo", "bar", "baz" ]
}
```

After reading the first one, the `Value` property is a string containing `"discrete"`, but after reading the second one (independently), `Value` will be `null` -- you must handle the array using the _Options.Abstractions_ package to deserialize via `Get<List<string>>`, or call `GetChildren()` to iterate over the contents (and in more complicated scenarios it may also be applicable to treat the value as a new `IConfigurationSection`). If you load _both_ versions of the JSON during startup (such as overlaying debug settings on production settings), then `Value` is populated _and_ the other data is also populated at the same time.

This package is expecting a single value to provide to a named argument, and there is no way to determine whether the contents of `Value` is the intended value or the complex data is the intended value.

bartelink:
I'm sure the readme and/or the Issues document all this stuff, but I really appreciate the detailed explanation. Will see if I can provide more useful comments given that context...

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
```diff
@@ -556,5 +559,43 @@ public void WriteToSubLoggerWithLevelSwitchIsSupported()
...
+
+        [Trait("Bugfix", "#103")]
+        [Fact]
+        public void MultipleArgumentValuesThrowsInvalidOperationException()
+        {
+            var jsonDiscreteValue = @"{
+                ""Serilog"": {            
+                    ""Using"": [""TestDummies""],
+                    ""WriteTo"": [{
+                        ""Name"": ""DummyRollingFile"",
```
bartelink: rolling file is deprecated, maybe example can start using `File` ?

MV10:
Huh, just learned the "View Changes" button also shows the old comments.

I'm also doing work in the SQL sink and that repo is all over the place in terms of style, but I agree, readability is very important. I'm also tempted to add _lots_ of comments but there doesn't seem to be much commenting in the Serilog codebase that I've seen.

I did consider splitting the line on the exception message, but even then it's still very long, I'm not sure much is to be gained there. I followed the example of a much [longer](https://github.com/serilog/serilog-settings-configuration/blob/dev/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs#L75) exception earlier in that file.

And it fits just fine on my ridiculous 3200x1800 screen! :grin:

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
```diff
@@ -556,5 +559,43 @@ public void WriteToSubLoggerWithLevelSwitchIsSupported()
 
             Assert.Equal(1, DummyRollingFileSink.Emitted.Count);
         }
+
+        [Trait("Bugfix", "#103")]
+        [Fact]
+        public void MultipleArgumentValuesThrowsInvalidOperationException()
```
bartelink: `InconsistentComplexVsScalarStateForAConfigurationNodeThrowsIOE`

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
```diff
@@ -556,5 +559,43 @@ public void WriteToSubLoggerWithLevelSwitchIsSupported()
...
+            // extension matching this exists (in theory an "object" argument could
+            // accept either value). ConfigurationReader should throw as soon as
+            // the multiple values are recognized; it will never attempt to locate
+            // a matching argument.
+
+            var ex = Assert.Throws<InvalidOperationException>(()
+                => ConfigFromJson(jsonDiscreteValue, jsonComplexValue));
+
+            Assert.Contains("Combined configuration sources", ex.Message);
+            Assert.Contains("pathFormat", ex.Message);
```
bartelink: use the `'` bounding in the message I added and assert on the full path (can you pin nesting/context from parent too, or is the as good as it gets) ?)

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
```diff
@@ -12,9 +12,12 @@ namespace Serilog.Settings.Configuration.Tests
 {
     public class ConfigurationSettingsTests
     {
-        static LoggerConfiguration ConfigFromJson(string jsonString)
+        static LoggerConfiguration ConfigFromJson(string jsonString, string secondJsonSource = null)
```
bartelink: could possibly make this `params` array, but rule of threee

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -221,6 +221,10 @@ void ApplyEnrichment(LoggerConfiguration loggerConfiguration, IReadOnlyDictionar
             IConfigurationArgumentValue GetArgumentValue(IConfigurationSection argumentSection)
             {
                 IConfigurationArgumentValue argumentValue;
+
+                if (argumentSection.Value != null && argumentSection.GetChildren().Any())
+                    throw new InvalidOperationException($"Combined configuration sources must result in a discrete value (string, int, etc.) or complex value (section, list, etc.), not both. Argument: {argumentSection.Path}");
```
bartelink: Suggest comment 
// Reject cases where incoming config is mangled such that there are both scalar and aggregate values for this section as invalid
Suggest some elements of following message 
"The configuration element: '{argumentSection.Path}', which is the result of combining multiple configuration inputs has an ambiguous set of values.\nPlease ensure inputs consistently supply either a scalar (int, string etc...) or a complex (section, list etc.) value for this argument section.");
"argument section" / "configuration element" / "scalar" are my terms and should only be used if they map to domain terms in serilog.settings.configuration or dotnetcore config stuff.

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
```diff
@@ -556,5 +559,43 @@ public void WriteToSubLoggerWithLevelSwitchIsSupported()
...
+
+            // These will combine into a ConfigurationSection object that has both
+            // Value == "C:\" and GetChildren() == List<string>. No configuration
+            // extension matching this exists (in theory an "object" argument could
+            // accept either value). ConfigurationReader should throw as soon as
+            // the multiple values are recognized; it will never attempt to locate
+            // a matching argument.
+
+            var ex = Assert.Throws<InvalidOperationException>(()
+                => ConfigFromJson(jsonDiscreteValue, jsonComplexValue));
```
bartelink: think `=>` shoudl go prev line ?

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
bartelink: hm, I see its used in the rest of the test codebase here; prob leave it

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -221,6 +221,10 @@ void ApplyEnrichment(LoggerConfiguration loggerConfiguration, IReadOnlyDictionar
             IConfigurationArgumentValue GetArgumentValue(IConfigurationSection argumentSection)
             {
                 IConfigurationArgumentValue argumentValue;
+
+                if (argumentSection.Value != null && argumentSection.GetChildren().Any())
```
bartelink: now you've explained it, this precondition check absolutely belongs here on its own and not nested

bartelink:
> I'm also doing work in the SQL sink and that repo is all over the place in terms of style, but I agree, readability is very important. I'm also tempted to add lots of comments but there doesn't seem to be much commenting in the Serilog codebase that I've seen.

One can def go overboard on comments, but for example the guard you're adding means nothing to me as I've never written code in the config system but for example can imagine myself ending up reading that code if I meet the exception. In the case of that exception message, if it can't just in time teach me by having a full article (brevity is important), then a comment can do half of the jon

>I did consider splitting the line on the exception message, but even then it's still very long, I'm not sure much is to be gained there. I followed the example of a much longer exception earlier in that file.

I try to:
```
"sentence. "
+ "And another thing"
```
as its a good forcing function for confronting the fact that you havent taken the time to write a short letter ;)

>And it fits just fine on my ridiculous 3200x1800 screen! 😁

:D I have a nice monitor here but 2 monitors in work are 1080 :( In general I look at github stuff in split view too as I need all the help I can get to remember how using short lines actually tends to cause better code in the end, even if it feels artificial!)

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
MV10: My thinking as well.

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
MV10: Not sure what you mean about `'` bounding? But the `Path` property does provide some additional context. I'll have to do some checking to see exactly how much, though, I think it changes depending on how you retrieve the section. On the other hand, `pathFormat` is as "deep" as you can go in this test data and still be processing anything related to the test target, so as-written it is arguably already a valid assertion.

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
MV10: And YAGNI... I can't think of a test scenario where 'n' config sources would do anything two sources can't represent.

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
MV10: I'm thinking:
`InconsistentComplexVsScalarArgumentValuesThrowsIOE`

MV10:
Great feedback. Much happier with the exception message in particular.

MV10:
Realized what you meant about `'` in the exception message and updated the test to assert against:

`'Serilog:WriteTo:0:Args:pathFormat'`

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
bartelink: I meant that the `=>` rarely starts a line in normal code and instead goes on prev line
i.e. the full `() =>` on prev line

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -221,6 +221,16 @@ void ApplyEnrichment(LoggerConfiguration loggerConfiguration, IReadOnlyDictionar
             IConfigurationArgumentValue GetArgumentValue(IConfigurationSection argumentSection)
             {
                 IConfigurationArgumentValue argumentValue;
+
+                // Reject configurations where an element has both scalar and complex
+                // values as a result of reading multiple configuration sources.
+                if (argumentSection.Value != null && argumentSection.GetChildren().Any())
+                    throw new InvalidOperationException(
+                        $"The value for the argument '{argumentSection.Path}' is assigned different value " +
```
bartelink: I keep thinking `ArgumentException` but I guess that's best reserved for actual method arguments.

MV10: That was my first inclination, too, but then I noticed IOE is used elsewhere in the code for similar "processing" errors.

nblumhardt:
:+1: great discussion here :-)

RE code style, unfortunately the Serilog org is pretty massive (a large number of sink repos) and consistency hasn't been our strong point. Mostly aiming for principle-of-least-surprise coding style; as much as I do love a good nitpick, there's a fair bit that comes down to taste and context. LGTM!

PR #120, opened 2018-07-15
Add support for .Destructure.AsScalar(Type scalarType) 

### LINKED ISSUE #115: Add support for .Destructure.AsScalar(Type scalarType)
For symmetry with *Serilog.Settings.AppSettings*, this should be supported : 

```json
 "Destructure": [
      { "Name": "AsScalar", "Args": { "scalarType": "Sample.SomeType, Sample" } }
    ],
```

This means : 
- supporting calls to `DestructuringConfiguration.AsScalar(...)` that is an instance method and not an extension method (therefore, needs some sort of *surrogate* method)
- supporting arguments of type `Type` when invoking methods

See serilog/serilog#1177 for implementation reference

tsimbalar:
Fixes #115 

- Support method arguments of type `Type` by converting from the type's qualified name
- Add support for (non-extension) method `Destructure.AsScalar`
- Add test to check that custom extension methods `Destructure.XXX()` can properly be invoked
- Refactor `ConfigurationReader` by extracting all the *surrogate method* handling to a separate class

### DIFF
--- a/sample/Sample/Program.cs
+++ b/sample/Sample/Program.cs
@@ -40,7 +40,7 @@ namespace Sample
                     new { TwentyChars = "0123456789abcdefghij" });
 
                 logger.Information("Destructure with max collection count:\n{@BigData}",
-                    new { TenItems = new string[] { "one", "two", "three", "four", "five", "six", "seven", "eight", "nine", "ten" } });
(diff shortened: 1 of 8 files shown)


MV10:
@tsimbalar love that change to move the Surrogate methods to their own home. That ConfigurationReader class was getting a bit tangled-up...

PR #136, opened 2018-09-29
fixes #63 - option to control assembly source

### LINKED ISSUE #63: Support loading Serilog Sinks at runtime
Hi there,

I wonder if serilog can load dlls (mainly sinks) dropped in the bin directory at runtime.
This feature will allow shipping a product with a default sink, while customers can later choose to use a different sink by dropping the dll in the bin directory and provide the right configuration for it.
Here is some information about the project:

- The project I am working on is a .net core project targeting .net 462 framework, so the AppDomain exists.
- The preserveCompilationContext is set to true. (false is not an option in this project)

I can see, the code in ConfigurationReader.GetSerilogConfigurationAssemblies() support assembly scanning  which is exactly what I need. However, there is no way that the ConfigurationLoggerConfigurationExtensions.ConfigurationSection(...) can pass a null DependencyContext value to trigger the assembly scanning, the logic around the DependencyContext is:

`dependencyContext ?? (Assembly.GetEntryAssembly() != null ? DependencyContext.Default : null`

This will only be evaluated to null if:

1. Dependency.Context.Default is null, which is not a suitable solution in my case as I need the preserveCompilationContext to be true.
2. Assembly.GetEntryAssembly() returns null, and this only happens when it's called from unmanaged code.
3. Write my own extension function that passes null to the ConfigurationSection(...). This cannot be achieved at the moment as the ConfigurationReader class is internal.

I would appreciate if you can point me to the right direction.

Thanks,
Ehab

MV10:
When the `DependencyContext` parameter is null, the `onNullDependencyContext` defines the interpretation -- either to scan in-memory assemblies (the default, when available) or scan DLL files in the working directory (optional, or the fallback if the default is not available). See #63 for discussion.


MV10:
Wonderful. All tests pass locally, but _two_ tests failed on AppVeyor -- the new one on `net452` for this PR, and failures for both sets of inputs for the untouched, pre-existing `ParameterlessSinksAreConfigured` (which I assume is probably quite old) on `netcoreapp2.0` only. Travis is fine with the new test, but fails the same older test on both inputs. (Looks like Travis doesn't run .NET Framework tests?)

Something is jacked up somewhere in the unit test pipeline. I've tried a bunch of local runs: sometimes they pass, sometimes they fail...

nblumhardt:
I like the usability of this one, but it seems like a grey area - do we really want to get into assembly scanning/loading here? It's historically been a very delicate and flaky part of .NET to get involved in. I'm not totally against it, but I'm also a bit nervous about supporting/debugging the failures that will inevitably bubble back to us, here....

Thoughts?

nblumhardt:
Hmm, just looking again, I might have misunderstood how this is implemented - I see it's just switching between `DependencyContext.Default` and `null`. Took a moment to get up to speed via the other thread :-)

src/Serilog.Settings.Configuration/ConfigurationLoggerConfigurationExtensions.cs
```diff
@@ -21,6 +21,22 @@
 
 namespace Serilog
 {
+    /// <summary>
+    /// Defines the interpretation of a null DependencyContext configuration parameter.
+    /// </summary>
+    public enum ConfigurationAssemblySource
```
nblumhardt: This could go under the `Serilog.Configuration` namespace without too much inconvenience.

src/Serilog.Settings.Configuration/ConfigurationLoggerConfigurationExtensions.cs
```diff
@@ -38,19 +54,23 @@ public static class ConfigurationLoggerConfigurationExtensions
...
-        /// default will be used.</param>
+        /// <param name="dependencyContext">The dependency context from which sink/enricher packages can be located.</param>
+        /// <param name="onNullDependencyContext">If dependency context is not supplied, either the platform default or DLL scanning may be used.</param>
         /// <returns>An object allowing configuration to continue.</returns>
         public static LoggerConfiguration Configuration(
             this LoggerSettingsConfiguration settingConfiguration,
             IConfiguration configuration,
-            DependencyContext dependencyContext = null)
+            DependencyContext dependencyContext = null,
+            ConfigurationAssemblySource onNullDependencyContext = ConfigurationAssemblySource.UseLoadedAssemblies)
```
nblumhardt: To make this a non-breaking change, this'd need to be added in an overload, rather than a new, defaulted parameter (which is still a binary-breaking change). If you consider that the value is never used in conjunction with `dependencyContext`, the overloaded method signature could be:

```csharp
        public static LoggerConfiguration Configuration(
            this LoggerSettingsConfiguration settingConfiguration,
            IConfiguration configuration,
            ConfigurationAssemblySource configurationAssemblySource)
```

The non-default value would prevent overload resolution conflicts.

MV10: Ah, much cleaner.

src/Serilog.Settings.Configuration/LoggerConfigurationSupport.cs
```diff
@@ -0,0 +1,32 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Serilog.Settings.Configuration
+{
+    /// <summary>
+    /// Defines how the package will identify the assemblies which are scanned for sinks and other Type information.
+    /// </summary>
+    public enum ConfigurationAssemblySource
```
nblumhardt: Nit: filename should be `ConfigurationAssemblySource.cs` under the `Settings/Configuration` subfolder

nblumhardt:
Just one nitpick comment, otherwise seems good-to-go, if we're satisfied this will sort out #63 👍

MV10:
No sweat, I debated about where to put that and whether to use a specific or general name.

Think this warrants documenting in the README? `DependencyContext` isn't explained so I figured both of these are sort of advanced usage, along the lines of "go read the code if you insist on playing with these"...

nblumhardt:
👍 I think we can deal with README updates down the line; pushing this through so we can assemble a final 3.0 dev build :-) 🎉

Issue #142, opened 2018-10-08
Extension methods with IConfiguration parameter are ignored


andriysavin:
When you define sink configuration extension method with `IConfiguration` parameter that method is ignored. At the same time if you use `IConfigurationSection` and specify value in config source, the method is discovered and the parameter is populated. Also, if default value (null) is specified for `IConfiguration` parameter all works as exepected as well. I believe the source of the problem lies in the code which filters methods and doesn't take into account `IConfiguration` parameter type: https://github.com/serilog/serilog-settings-configuration/blob/51a4fca7cf125adc717b657afa98ea8db2fddaea/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs#L351

nblumhardt:
Thanks for the report, sounds like something was overlooked 👍

MV10:
I don't think it's a bug in the config package.. The SQL sink relies on this to find a named connectionstring in a section outside of the Serilog-specific section. Using the dev version of the SQL sink and the 3.0.0 release of the config package, this code and config works and is probably the simplest way to reproduce this without writing your own sink:

```csharp
var appConfig = new ConfigurationBuilder()
    .SetBasePath(Directory.GetCurrentDirectory())
    .AddJsonFile(path: "appsettings.json")
    .Build();

var logConfig = new LoggerConfiguration();
logConfig.ReadFrom.Configuration(appConfig);
Log.Logger = logConfig.CreateLogger();
Serilog.Debugging.SelfLog.Enable(Console.Error);

Log.Information("An Information message");

Log.CloseAndFlush();
```

```json
{
  "ConnectionStrings": {
    "DevTest": "Data Source=(LocalDB)\\MSSQLLocalDB;Initial Catalog=SerilogTest;Integrated Security=true"
  },
  "Serilog": {
    "Using": [ "Serilog.Sinks.Console" ],
    "WriteTo": [
      { "Name": "Console" },
      {
        "Name": "MSSqlServer",
        "Args": {
          "connectionString": "DevTest",
          "tableName": "Issue81",
          "autoCreateSqlTable":  true,
          "columnOptionsSection": {
            "customColumns": [
              {
                "ColumnName": "EventType",
                "DataType": "int"
              },
              {
                "ColumnName": "Release",
                "DataType": "varchar",
                "DataLength": 32
              }
            ]
          }
        }
      }
    ]
  }
}
```

andriysavin:
[...]

PR #144, opened 2018-10-13
Do not access "Serilog" sub-section when calling ReadFrom.ConfigSection / properly populate `IConfiguration` in configuration methods

### LINKED ISSUE #143: ConfigurationSection() adds sub-section in version 3.0.0; configurable default section name
I've been using the `ReadFrom.ConfigurationSection()` extension method solely for reading the logger configuration from a configuration section with a name that differs from the hard-coded default (e.g. "Logging" instead of "Serilog").

Version 3.0.0 no longer uses the provided configuration section directly, but instead calls `GetSection(DefaultSectionName)`, adding a hard-coded sub-section. This change was seemingly introduced with commit 80b2ec98c56394b9f6901a33c7db64af62f4c652 or one thereafter. I am not sure whether this is on purpose, but using the provided configuration section directly as in version 2.6.1 and earlier seems more logical and also fits the documented behaviour of the extension method.
After having a quick look at the code path in the debugger, it seems that probably just the wrong constructor overload for `ConfigurationReader` is called in the 3.0.0 NuGet package assembly.

Regardless of the descrived issue, it would be great if the default section name could be configured (or provided as an optional parameter). If that were possible, the recommended `ReadFrom.Configuration()` extension method could be used instead if the only difference is just a different section name.

tsimbalar:
fixes #143 where we are trying to read a subsection of config section passed as parameter

This is a regression introduced in v3.0.0

The `ConfigurationReader` had several constructor overloads, one of them accepting a `IConfigurationSection` instead of a `IConfiguration`... but the `IConfigurationSection` one was not being invoked by the `ReadFrom.ConfigSection()` extension methods.

Making that constructor `public` allows the section to be passed, but it also means the `ConfigurationReader` no longer has access to the *full* `Configuration` when that overload is used, so the behavior of *Serilog.Settings.Configuration* with config methods accepting `IConfiguration` parameters may be subtly broken. 

I am wondering if to properly solve that we should : 
- mark `ReadFrom.ConfigSection` as obsolete
- add a new `ReadFrom.ConfigSection(Configuration, sectionName:string)` 

that is, we would still pass the full `Configuration` object, but look for a given section within ... 

Any lights on this topic are welcome, I'm not 100% up-to-date with how *Microsoft.Extensions.Configuration* works :)

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -35,9 +35,10 @@ namespace Serilog.Settings.Configuration
 
         // Generally the initial call should use IConfiguration rather than IConfigurationSection, otherwise
         // IConfiguration parameters in the target methods will not be populated.
-        ConfigurationReader(IConfigurationSection configSection, DependencyContext dependencyContext)
+        public ConfigurationReader(IConfigurationSection configSection, DependencyContext dependencyContext)
         {
-            _section = configSection ?? throw new ArgumentNullException(nameof(configSection));
+            _configuration = configSection ?? throw new ArgumentNullException(nameof(configSection));
+            _section = configSection;
             _dependencyContext = dependencyContext;
             _configurationAssemblies = LoadConfigurationAssemblies();
         }
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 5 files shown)


src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -35,9 +35,10 @@ public ConfigurationReader(IConfiguration configuration, DependencyContext depen
 
         // Generally the initial call should use IConfiguration rather than IConfigurationSection, otherwise
         // IConfiguration parameters in the target methods will not be populated.
-        ConfigurationReader(IConfigurationSection configSection, DependencyContext dependencyContext)
+        public ConfigurationReader(IConfigurationSection configSection, DependencyContext dependencyContext)
         {
-            _section = configSection ?? throw new ArgumentNullException(nameof(configSection));
+            _configuration = configSection ?? throw new ArgumentNullException(nameof(configSection));
```
MV10: It was probably an oversight to leave `public` off this, but the comment mentions that `IConfiguration` parameters won't be populated in target methods, and populating `_configuration` isn't likely to help. The only place this matters that I'm aware of (so far) is in the SQL sink where the full config is used to look for named connection strings (which would be a config section external to the Serilog section, whatever name it might use).

I'd have to study this and think about the issue more but offhand I can't think of any other problems fixing it this way. I'd just leave `_configuration` null (without looking at the code, I assume the package would treat a null as not having an `IConfiguration` available to populate in the target method).

tsimbalar:
Ok, so it turns out that, while implementing an explicit error message for cases where calling `ReadFrom.ConfigSection` and trying to call configuration methods expecting an `IConfiguration`, I could reproduce #142 ...

This PR therefore fixes both #143 and #142 , and introduces tests for those cases.

Unfortunately, because we rely on a `static` `ConfigurationReader._configuration` field, our tests are no longer deterministic (the test `LoggerConfigurationExtensionsTests.ReadFromConfigurationSectionThrowsWhenTryingToCallConfigurationMethodWithIConfigurationParam` passes when run alone, but not anymore where the whole test suite is run). 

I think the long term fix would be to get rid of that `static` state, but it may be better to release a *v3.0.1* bugfix first ? 

@nblumhardt @MV10 any opinion on the subject ?

nblumhardt:
Love the added test coverage :-) 👍 

+1 from me, seems like getting this to dev and asking for some confirmation on the original report would be the best way to move it forward. @MV10 ?

PR #146, opened 2018-10-15
Bugfix release v3.0.1

### LINKED ISSUE #142: Extension methods with IConfiguration parameter are ignored

### LINKED ISSUE #143: ConfigurationSection() adds sub-section in version 3.0.0; configurable default section name

tsimbalar:
We have a few bugs in v3.0.0. It would be nice to publish a release with the fixes quickly :) 

- fixes #142 : `IConfiguration` parameters not populated
- fixes #143 : `ReadFrom.ConfigSection` expects nested section under `Serilog` section

See discussion in https://github.com/serilog/serilog-settings-configuration/pull/144#issuecomment-429574928


tsimbalar:
When https://github.com/serilog/serilog-settings-configuration/pull/149 gets merged, we may be in a good spot to publish a bugfix release v3.0.1, what do you think ? 

@nblumhardt @MV10 @merbla

MV10:
I'll be on the road and email-only until Sunday, so I can't look at anything in-depth. Sounds like the right thing to do, though.

nblumhardt:
Awesome :-) 🎉

PR #158, opened 2019-01-04
Update Nuget package license format to new format

tsimbalar:
`licenseUrl` is now deprecated so we use a `license expression`

Tracked in https://github.com/serilog/serilog/issues/1265

This properly populate the license information in the nuget package : 

image


skomis-mm:
👍  

Would be useful to also check license type=`File` instead (because we have `https://github.com/serilog/serilog-settings-configuration/blob/dev/LICENSE` file) for having less reasons to sync them (unlikely though :)).

tsimbalar:
I think which approach we choose does not matter too much as long as we are consistent across the Serilog repos :)

Issue #157, opened 2018-12-16
Also look for dependencies in the `bin` folder


tsimbalar:
As pointed out by https://github.com/serilog/serilog-settings-configuration/issues/156#issuecomment-446879312 .

This will fix an issue here dependencies are automatically discovered in a .NET Framework console app (where the `.exe` and its dependencies typically live in the same folder) but not in .NET Framework ASP.NET apps (where dependencies are copied to the `bin` folder)

tsimbalar:
This should get easier to implement once the refactorings in https://github.com/serilog/serilog-settings-configuration/pull/151 are merged

Issue #156, opened 2018-12-10
Sublogger not working in web host


gautelo:
When using a web host subloggers don't seem to work.

I've created a very stripped down version of what we're doing and uploaded that as a [reference repo](https://github.com/gautelo/bug-serilog-sublogger-web) in the hopes that you might point out what I'm doing wrong, or figure out why it's failing.

As you can see I have a web host and a console host. They both use the same logging configuration, yet the sublogger only work for the console host. The interresting bit is that the straight file logger works for both hosts, so it's not that serilog is not working at all, it's just that the subloggers aren't working.

I've also included an alternate configuration option for configuring through code instead of through settings. Swap between `SerilogConfig.RegisterBySettings` and `SerilogConfig.RegisterByCode` to see how using `SerilogConfig.RegisterByCode` causes the sublogger to work in both hosts.

Note: We're in a bit of a crossover between older and newer tech which might be a reason why this does not work, though I don't see why it would cause an issue seing as it works for `SerilogConfig.RegisterByCode`.

tsimbalar:
Thanks a million for the repro ! You've obviously spent some time preparing it, so it feels only fair to spend some time reviewing it ;) 

I don't have the proper environment right now to actually run it, but here are my 2 cents :) 

> Swap between SerilogConfig.RegisterBySettings and SerilogConfig.RegisterByCode to see how using SerilogConfig.RegisterByCode causes the sublogger to work in both hosts.

this definitely makes me think there is something fishy about the configuration file somehow, so I'll focus on that for now :) ...

I see your json config file for the "sub-logger" looks like this : 

```json
{
  "Serilog": {
    "MinimumLevel": "Information",
    "Enrich": [
      "WithThreadId"
    ],
    "Using": {
      "File": "Serilog.Sinks.File"
    },
    "WriteTo": {
      "MainFile": {
        "Name": "Logger",
        "Args:configureLogger:WriteTo:File": {
          "Name": "File",
          "Args": {
            "path": "%SERILOG_OUTPUT_PATH%\\ConfigSub-.log",
            "outputTemplate": "{Timestamp:HH:mm:ss.fff} {Level:u3} {ThreadId,3} {SourceContext} :> {Message}{NewLine}{Exception}",
            "shared": true,
            "flushToDiskInterval": "00:00:03",
            "rollingInterval": "Day",
            "retainedFileCountLimit": 31
          }
        }
      }
    }
  }
}
```

and 2 things caught my attention : 

[...]

Issue #150, opened 2018-10-16
ConfigurationReader in Android throw exceptions.


Lukiya:
ConfigurationReader.cs line 312.

[query = from outputAssemblyPath in System.IO.Directory.GetFiles(AppDomain.CurrentDomain.BaseDirectory, "*.dll")](https://github.com/serilog/serilog-settings-configuration/blob/dev/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs#L312)

**_AppDomain.CurrentDomain.BaseDirectory_** cannot be used in Android platform, it always return null.
Due to this, a System.ArgumentNullException will be thrown out.

MV10:
I'm betting this would also fail on iOS (though I have no way to test this).

Lukiya:
Fortunately, iOS doesn't have this issue. Here is how I get the base directory in android.
I pass the main entry class type into my app, and use that type to get its location, it works pretty well.

```
private string GetBinDirectory(Type entryPointType = null)
{
    if (null!= entryPointType)
    {
        var binDirectory = Path.GetDirectoryName(entryPointType.Assembly.Location);
    }
    else
    {
        var binDirectory = AppDomain.CurrentDomain.BaseDirectory;
    }
    return binDirectory;
}
```

tsimbalar:
This should get easier to implement once the refactorings in PR https://github.com/serilog/serilog-settings-configuration/pull/151 are merged.

Issue #122, opened 2018-07-19
.NET Framework: Configuring Enrichers from json does not work


aFuerst:
**Does this issue relate to a new *feature* or an existing *bug*?**
- [X] Bug
- [ ] New Feature

**What version of Serilog is affected?  Please list the related NuGet package.**
Serilog, Version=2.0.0.0
Serilog.Enrichers.Process, Version=2.0.0.0
Serilog.Enrichers.Thread, Version=2.0.0.0
Serilog.Enrichers.Environment, Version=2.0.0.0
Serilog.Settings.Configuration, Version=2.6.1.0

**What is the target framework and operating system? See [target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) & [net standard matrix](https://docs.microsoft.com/en-us/dotnet/standard/net-standard).**

- [ ] netCore 2.0
- [ ] netCore 1.0
- [X] 4.7
- [ ] 4.6.x
- [ ] 4.5.x

**Please describe the current behavior?**
Enrichers configured via json configuration are not applied. 
The created logger has an empty enricher.

**Please describe the expected behavior?**
The enrichers specified in the json config should be applied to the created logger and the data they are supposed to add gets put in the logs

**If the current behavior is a bug, please provide the steps to reproduce the issue and if possible a minimal demo of the problem**
This is the config section I have
```json
"Serilog": {
	"Enrich": [ "FromLogContext", "WithMachineName", "WithProcessId", "WithThreadId", "WithExceptionDetails" ],
	"MinimumLevel": "Verbose",
	"WriteTo": [
		{
			"Name": "Trace",
			"Args": {
				"formatter": "Serilog.Formatting.Json.JsonFormatter, Serilog"
			}
		}
	],
	"Using": [ "Serilog.Sinks.Trace" ]
}
```
We're using the full .NET Framework. 
The logger creation looks like this. We load it entirely from the IConfiguration object.
```csharp
var logger = new LoggerConfiguration()
	.ReadFrom.Configuration(config)
	.CreateLogger();
```

merbla:
@aFuerst are you missing a using in the settings for `Serilog.Enrichers.Environment?

```
"Using": [ "Serilog.Sinks.Trace", "Serilog.Enrichers.Environment" ]
```

aFuerst:
@merbla That was it. Thank you!
I didn't realize I needed to add the namespaces to the `Using` since you didn't do that in the example in the ReadMe

[...]

PR #159, opened 2019-01-04
DllScanningAssemblyFinder fixes (#157, #150, #122, #156)

skomis-mm:
- `ArgumentNullException` fix when `AppDomain.CurrentDomain.BaseDirectory` is null
- additional paths probing:  `AppDomainSetup.PrivateBinPath`, current assembly location path when `BaseDirectory` is null

Fixes [#150](https://github.com/serilog/serilog-settings-configuration/issues/150), [#157](https://github.com/serilog/serilog-settings-configuration/issues/157), [#122](https://github.com/serilog/serilog-settings-configuration/issues/122), [#156](https://github.com/serilog/serilog-settings-configuration/issues/156)

### DIFF
--- a/test/Serilog.Settings.Configuration.Tests/DllScanningAssemblyFinderTests.cs
+++ b/test/Serilog.Settings.Configuration.Tests/DllScanningAssemblyFinderTests.cs
@@ -9,13 +9,17 @@ namespace Serilog.Settings.Configuration.Tests
 {
     public class DllScanningAssemblyFinderTests : IDisposable
     {
+        const string BinDir1 = "bin1";
+        const string BinDir2 = "bin2";
+        const string BinDir3 = "bin3";
+
         readonly string _privateBinPath;
 
         public DllScanningAssemblyFinderTests()
         {
-            var d1 = GetOrCreateDirectory("bin1");
-            var d2 = GetOrCreateDirectory("bin2");
-            var d3 = GetOrCreateDirectory("bin3");
+            var d1 = GetOrCreateDirectory(BinDir1);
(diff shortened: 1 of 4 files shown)


test/Serilog.Settings.Configuration.Tests/DllScanningAssemblyFinderTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Serilog.Settings.Configuration.Tests
+{
+    public class DllScanningAssemblyFinderTests : IDisposable
+    {
+        readonly string _privateBinPath;
+
+        public DllScanningAssemblyFinderTests()
+        {
+            var d1 = GetOrCreateDirectory("bin1");
```
tsimbalar: should `"bin1"`, `"bin2"` and `"bin3"` be moved to constants instead of using *magic strings* multiple time ?

skomis-mm:
@tsimbalar there is not much difference: `DependencyContext` can be used for both, `CurrentDomain.BaseDirectory` can be used for both too. 
The difference for classic web applications is additional probing path (`bin` directory via setting `AppDomainSetup.PrivateBinPath` which can be configured only for the Full Framework.

test/Serilog.Settings.Configuration.Tests/DllScanningAssemblyFinderTests.cs
skomis-mm: Okay

tsimbalar:
LGTM ! :shipit:

PR #169, opened 2019-03-20
Configuration File Path Edited

hbergun:
Configuration file path does not work as in documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ public class Program
     public static void Main(string[] args)
     {
         var configuration = new ConfigurationBuilder()
+	    .SetBasePath(AppContext.BaseDirectory.Substring(0, AppContext.BaseDirectory.IndexOf("bin")))
             .AddJsonFile("appsettings.json")
             .Build();
 


nblumhardt:
Thanks for sending the PR. The correct path here will depend on the app that's being written (not all paths have `bin` in them); the [documentation](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-2.2#json-configuration-provider) all seems to use `Directory.GetCurrentDirectory()` so we could follow that convention?

```
    .SetBasePath(Directory.GetCurrentDirectory())
```

hbergun:
> Thanks for sending the PR. The correct path here will depend on the app that's being written (not all paths have `bin` in them); the [documentation](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-2.2#json-configuration-provider) all seems to use `Directory.GetCurrentDirectory()` so we could follow that convention?
> 
> ```
>     .SetBasePath(Directory.GetCurrentDirectory())
> ```

yes that is correct for on kestrel server.if we use Visual Studio(Not:VsCode) and runing on IIS server
```C#
Directory.GetCurrentDirectory();
```
The code line above gives the location of the IIS and this will cause the error

PR #176, opened 2019-05-03
added test to show how to filter child contexts

dzmitry-lahoda:


test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
```diff
@@ -194,6 +194,39 @@ public void TestMinimumLevelOverrides()
...
+                .CreateLogger();
+
+            log.Write(Some.DebugEvent());
+            Assert.Null(evt);
+
+            var custom = log.ForContext("SourceContext", "System.Threading.Tasks.Task<42>");
+            custom.Write(Some.DebugEvent());
+            Assert.NotNull(evt);
+
+            var systemThreadingLogger = log.ForContext<System.Threading.Tasks.Task>();
```
tsimbalar: maybe you would need to reset `evt` back to `null` to make the last assertion really valid ?

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
```diff
@@ -194,6 +194,39 @@ public void TestMinimumLevelOverrides()
...
+            LogEvent evt = null;
+
+            var log = ConfigFromJson(json)
+                .WriteTo.Sink(new DelegatingSink(e => evt = e))
+                .CreateLogger();
+
+            log.Write(Some.DebugEvent());
+            Assert.Null(evt);
+
+            var custom = log.ForContext("SourceContext", "System.Threading.Tasks.Task<42>");
```
tsimbalar: is there any reason to not use 

```cs
log.ForContext<SystemThreading.Tasks.Task<Int32>>();
```
(or something similar)

dzmitry-lahoda: We use `MyGame.Room<489414134>`.  So not type, but value.

dzmitry-lahoda: There is seems issue in `serilog-sinks-file` as it does not work with child matching.

tsimbalar: oh ok, fair enough :P 
You could be using `Serilog.Core.Constants.SourceContextPropertyName` instead of `"SourceContext"` then ;)

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
dzmitry-lahoda: will check.

dzmitry-lahoda:
This file has broken editorconfig. All JSON inside are broken. I have not fixed it. Too much noise. Issue is not relevant to my pull.

dzmitry-lahoda:
https://github.com/serilog/serilog-sinks-file/issues/100

nblumhardt:
Thanks @dzmitry-lahoda :+1:

PR #179, opened 2019-05-23
Added missing null checks for settingConfiguration

marcjohne:
Added some missing null checks for `settingConfiguration` arguments.


nblumhardt:
Thanks!

Issue #167, opened 2019-03-17
Is it possible to specify value for named parameters of constructor


vhatsura:
Hey, guys!

In some case I need to specify named parameters of constructor (in most cases, it's needed for custom formatters).
Let's see it on a simple example.
To use custom formatter with Console sink we can use the following configuration:

```json
{
  "Serilog": {
    "WriteTo": [
      {
        "Name": "Console",
        "Args": {
          "formatter": "JsonValueFormatter "
        }
      }
    ]
  }
}
```

Is it possible to somehow in above example to specify [`typeTagName`](https://github.com/serilog/serilog/blob/9a46e68d3e224b965f4ac2665fd379ae9568d6b4/src/Serilog/Formatting/Json/JsonValueFormatter.cs#L39) parameter of constructor?  

If it isn't possible to do, I will appreciate for ability to implement such functionality by myself.

tsimbalar:
Hi Vadim,

no it is not currently supported :-/ 
There has been a similar request here, with a proposed workaround if that works for you : https://github.com/tsimbalar/serilog-settings-comparison/issues/1

vhatsura:
Hi, @tsimbalar.

Thanks for the link. The proposed workaround is interisting and can be helpful, however not in all scenario.

As an example, for [ExceptionAsObjectJsonFormatter from Serilog.Formatting.Elasticsearch package](https://github.com/serilog/serilog-sinks-elasticsearch/blob/f38b376b7fb629c07c45a8573823d86fd858a4fe/src/Serilog.Formatting.Elasticsearch/ExceptionAsObjectJsonFormatter.cs#L46) it can be hard to precreate formatters with all possible combinations.

PR #188, opened 2019-06-25
Support of constructor parameters

vhatsura:
related to #167 

Hey, guys! It's an early draft pull request with implementation of possiblity to create object via constructors with parameters. I will be appreciate if you can take a look and add your feedback

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ObjectArgumentValue.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ObjectArgumentValue.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Reflection;
 
 using Serilog.Configuration;
+using System.Linq;
 
 namespace Serilog.Settings.Configuration
 {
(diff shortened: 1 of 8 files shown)


nblumhardt:
Thanks! Short on time but will try to loop back around. What was the situation in which you found yourself needing this? (I've seen several :+1:)

vhatsura:
In my case I use [`ExceptionAsObjectJsonFormatter`](https://github.com/serilog/serilog-sinks-elasticsearch/blob/729b506c0a13aa44ff17b5029291cc79e8cb3527/src/Serilog.Formatting.Elasticsearch/ExceptionAsObjectJsonFormatter.cs#L29) from `Serilog.Formatting.Elasticsearch`. The formatter has a lot of arguments and sometimes we need to override them. This is why ability to configure constructor arguments in configuration is nice to have

julealgon:
@nblumhardt , what's the status on this? I just stumbled upon this problem today when introducing file logging in one of our solutions, and realizing I needed to set the `renderMessage` argument on `JsonFormatter` to `true`.

Would be really useful to have first-class support for configuring complex/abstract instances with parameters in config files.

tsimbalar:
@julealgon for now, you might want to use a workaround such as the one described in 
https://github.com/serilog/serilog-sinks-console/issues/28

or follow indications in https://github.com/tsimbalar/serilog-settings-comparison/blob/master/docs/README.md#interfaces-and-abstract-classes , and make a `Formatter` configured the way you want it either as : 
- a custom class with a default constructor
- a static property on your own type

tzographos:
> @julealgon for now, you might want to use a workaround such as the one described in
> [serilog/serilog-sinks-console#28](https://github.com/serilog/serilog-sinks-console/issues/28)

Hello @tsimbalar. How's this solving the problem where there is a need to configure a TextFormatter with constructor arguments (non-parameterless constructor)?

tsimbalar:
@tzographos you're right, there is nothing about this in https://github.com/serilog/serilog-sinks-console/issues/28 ... no idea what I meant back then 🤔

skomis-mm:
Ok, I've just created the [new one](https://github.com/serilog/serilog-settings-configuration/pull/281) with additional features. Feel free to comment/review :)

PR #191, opened 2019-07-25
Feature/metadata assembly finder

anAgent:
New Configuration extension to resolve the Azure Functions Assembly's DependencyContext by the specified Assembly - looking at the "dep.json" files that is generated on build.  Added example solution to demonstrate it's usaged.


nblumhardt:
Thanks for the comprehensive PR! This will take some time to digest/review; hopefully someone watching this repository with the requisite knowledge will be able to pick it up soon.

anAgent:
There is one flaw that my team mate brought to my attention in that this solution will truly only work for Azure Functions rather than all **netcoreapp2.x** apps. 

I think to make it actually work, I will need to make changes to the code so that it looks in the right directory for the "*.deps.json" file. Here's why:

With the context being **.\Debug\netcoreapp2.x** folder for a standard NetCoreApp2.x, the file structure for the .deps.json looks like the following.
image

```cmd
    appsettings.json
    Sample.deps.json
    Sample.dll
    Sample.pdb
    Sample.runtimeconfig.dev.json
    Sample.runtimeconfig.json
    Serilog.Settings.Configuration.dll
    Serilog.Settings.Configuration.pdb
    Serilog.Settings.Configuration.xml
```

For an Azure Function, the output looks like the following. Please take note that the Azure function doesn't run at the root of "netcoreapp2.0", but rather a sub-directory of the name of the function. In the following, it's "**.\netcoreapp.2.0\MyFunction**"

image

```cmd
netcoreapp2.0

│   appsettings.json
│   FunctionApp.deps.json
│   host.json
│   local.settings.json
│   tree.txt
│
├───bin
│   │   FunctionApp.dll
│   │   FunctionApp.pdb
│   │   Serilog.dll
│   │   Serilog.Enrichers.Environment.dll
│   │   Serilog.Settings.Configuration.dll
│   │   Serilog.Settings.Configuration.pdb
│   │   Serilog.Settings.Configuration.xml
│   │   Serilog.Sinks.ApplicationInsights.dll
│   │   Serilog.Sinks.Console.dll
│   │
│   └───runtimes
│       ├───unix
│       │   └───lib
│       │       └───netstandard1.3
│       │               System.Text.Encoding.CodePages.dll
│       │
│       └───win
│           └───lib
│               └───netstandard1.3
│                       System.Text.Encoding.CodePages.dll
│
└───MyFunction <<<<<< This is the context that the function is run under. >>>>>>
        function.json
```
Finally, with all that out of the way. I'm not sure about the best solution. 

Would you like an additional _Assembly Finder_ that works for non-Azure Function Apps or to modify the code I put in place to try and resolve where the .deps.json file resides. This is just a quick wack at  the code change. It's pretty assumptive that the fist attempt to locate the file is expecting it to be in the parent directory. However, it will stop if it can't resolve it...

```csharp
private DependencyContext GetDependencyContext() {
    var rootBinDirectory = Path.GetDirectoryName(_assembly.Location);
    var fileName = Path.ChangeExtension(_assembly.ManifestModule.Name, ".deps.json");
    var dependencyJsonFile = Path.Combine(
        rootBinDirectory ?? throw new InvalidOperationException("Unable to locale Assembly Location."),
        fileName
    );

    // If the Assembly is for an Azure Function, then the file will be located up one directory from the DLL.
    if (!File.Exists(dependencyJsonFile)) {
        dependencyJsonFile = Path.Combine(Directory.GetParent(rootBinDirectory).FullName, fileName);
    }

    if (!File.Exists(dependencyJsonFile)) {
        throw new NullReferenceException($"Unable to locate {dependencyJsonFile}.");
    }

    return new DependencyContextJsonReader()
        .Read(File.OpenRead(dependencyJsonFile));
}
```

Thanks!

skomis-mm:
Hi @anAgent ,

Thanks for the time and effort. I think we can leave things as is here to be Azure Functions agnostic and utilize current extension point passing custom `DependencyContext` or use `Using` section. I have added solution in corresponding issue.

PR #201, opened 2019-11-19
Add support for custom types in arrays and custom collections

sungam3r:
Also:

- add new test platforms and unify msbuild conditional properties
- minor code cleanup
- PackageIcon instead of PackageIconUrl
- fix for simplified syntax on netcore3.0
- fix SinkWithConfigurationBindingArgument test (wrong argument type)


test/Serilog.Settings.Configuration.Tests/Support/JsonStringConfigSource.cs
```diff
@@ -45,7 +45,7 @@ public JsonStringConfigProvider(string json) : base(new JsonConfigurationSource
 
             public override void Load()
             {
-                Load(StringToStream(_json));
+                Load(StringToStream(_json.Fix()));
```
sungam3r: netcore3.0 System.Text.Json issue

test/TestDummies/DummyLoggerConfigurationExtensions.cs
```diff
@@ -63,13 +63,20 @@ public static LoggerConfiguration WithDummyThreadId(this LoggerEnrichmentConfigu
 
         public static LoggerConfiguration DummyRollingFile(
             this LoggerSinkConfiguration loggerSinkConfiguration,
-            List<string> objectBinding,
+            List<Binding> objectBinding,
```
sungam3r: @nblumhardt In fact, a test using this method should have failed because it sets json incorrectly:
```
 var json = @"{
                ""Serilog"": {            
                    ""Using"": [""TestDummies""],
                    ""WriteTo"": [{
                        ""Name"": ""DummyRollingFile"",
                        ""Args"": {""pathFormat"" : ""C:\\"",
                                   ""objectBinding"" : [ { ""foo"" : ""bar"" }, { ""abc"" : ""xyz"" } ] }
                    }]        
                }
            }";
```

`objectBinding` is not a array of string, it is array of objects. I came across this error when adding support for arrays 
 of custom types.

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -262,6 +237,31 @@ string GetSectionName(IConfigurationSection s)
             }
         }
 
+        internal static IConfigurationArgumentValue GetArgumentValue(IConfigurationSection argumentSection, IReadOnlyCollection<Assembly> configurationAssemblies)
```
sungam3r: it began to be used in another place

src/Serilog.Settings.Configuration/Settings/Configuration/ObjectArgumentValue.cs
```diff
@@ -47,8 +47,72 @@ public object ConvertTo(Type toType, ResolutionContext resolutionContext)
...
+                {
+                    var argumentValue = ConfigurationReader.GetArgumentValue(configurationElements[i], _configurationAssemblies);
+                    var value = argumentValue.ConvertTo(elementType, resolutionContext);
+                    result.SetValue(value, i);
+                }
+
+                return result;
+            }
+
+            bool TryCreateContainer(out object result)
```
sungam3r: If false fallbacks to ms config binding implementation.

sungam3r:
@nblumhardt Could you please find the time to see this PR?

test/Serilog.Settings.Configuration.Tests/Support/Extensions.cs
```diff
@@ -8,5 +8,15 @@ public static object LiteralValue(this LogEventPropertyValue @this)
         {
             return ((ScalarValue)@this).Value;
         }
+
+        // netcore3.0 error:
+        // Could not parse the JSON file. System.Text.Json.JsonReaderException : ''' is an invalid start of a property name. Expected a '"'
```
nblumhardt: Where do these come from? (Would be better to fix the source JSON if possible).

nblumhardt:
Thanks for the nudge @sungam3r - I'm maxed out currently but will try to get back to this shortly, unless another of the maintainers of this repo gets here first :-)

Changes seem good. Just for context, is there a sink we fail to configure currently because of this?

test/Serilog.Settings.Configuration.Tests/Support/Extensions.cs
sungam3r: Source json does not need to be changed. There are many tests that test the simplified syntax. The problem is not in the tests, but in the fact that the new serializer, in principle, does not support the simplified syntax.

sungam3r:
I needed this in my own project. I wanted to configure destructuring policy (not a sink) accepting a list of types (restrictions).

test/Serilog.Settings.Configuration.Tests/Support/Extensions.cs
nblumhardt: :+1:

Perhaps `ToValidJson()` would be more descriptive here? Since it's exploiting some fairly questionable JSON parsing on the earlier platform, we could get rid of the conditional and just use the same method on all platforms?

test/Serilog.Settings.Configuration.Tests/Serilog.Settings.Configuration.Tests.csproj
```diff
@@ -12,14 +12,15 @@
...
+    <LangVersion>latest</LangVersion>
+    <TargetFrameworks>netcoreapp3.0;netcoreapp2.2;netcoreapp2.0;net452</TargetFrameworks>
     <AssemblyName>Serilog.Settings.Configuration.Tests</AssemblyName>
     <AssemblyOriginatorKeyFile>../../assets/Serilog.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(TargetFramework)' == 'net452'">
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net451' or '$(TargetFramework)' == 'net452'">
```
nblumhardt: `net451` isn't in `<TargetFrameworks>`

src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
```diff
@@ -1,10 +1,11 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Description>Microsoft.Extensions.Configuration (appsettings.json) support for Serilog.</Description>
     <VersionPrefix>3.1.1</VersionPrefix>
+    <LangVersion>latest</LangVersion>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>netstandard2.0;net451;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net451;net452;net461</TargetFrameworks>
```
nblumhardt: What's the reason for the additional target framework here?

test/Serilog.Settings.Configuration.Tests/Support/Extensions.cs
sungam3r: Breaking changes?

nblumhardt: Not that I'm aware of; the code is in the test project (so the changes are on the .NET Core side)

sungam3r: I do not understand. The simplified syntax should continue to work. Tests for it should remain, it makes no sense to rewrite them.

> just use the same method on all platforms?

What does this mean?

src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
sungam3r: I added this target because I saw references to it in the conditions.

test/Serilog.Settings.Configuration.Tests/Support/Extensions.cs
sungam3r: done in #203

test/Serilog.Settings.Configuration.Tests/Serilog.Settings.Configuration.Tests.csproj
sungam3r: done in #203

sungam3r:
I accidentally made a rebase on the wrong branch. So I created a new PR - #203

PR #203, opened 2019-12-07
Add support for custom types in arrays and custom collections

sungam3r:
Also:
- add new test platforms and unify msbuild conditional properties
- minor code cleanup
- PackageIcon instead of PackageIconUrl
- fix for simplified syntax on netcore3.0
- fix SinkWithConfigurationBindingArgument test (wrong argument type)

relates to #201

PR #205, opened 2020-01-18
Update to lts 31

tebeco:
This PR follow the logic of both :
https://github.com/serilog/serilog-aspnetcore/pull/165
https://github.com/serilog/serilog-extensions-hosting/pull/17

What this PR change :
* Add `global.json`
* Add `netcoreapp3.1` as a `TargetFramework` (Not the best way of resolution but it's still better than doing nothing)
* Update AppVeyor image to Vs2019
* Remove linter warning in `powershell` script
* Use `PackageIcon` as `PackageIconUrl` is being deprecated
* Always Resolve latest dependencies at build time for `Serilog`

## Thoughts :
As said in this PR :
https://github.com/serilog/serilog-extensions-hosting/pull/17
```
we could also just stop support for 2.x because 3.1 is LTS <=== This would require less complexity here, and customer using runtime/Microsoft.Exntesion.* would just need to pin this dependency to the current release
Note that the fact that a customer does not want to update it own dependency is driven to "pin down" it dependencies as they need evolve.
I have no idea What is Serilog stance on that but this is what Major is for ;)

For now i've tried to do this by using cross targeting, but as said the proper solution for this lib is to drop support for any Microsoft.Extensions.xxx 2.x as 3.1 is LTS since December 2019
```

## Side note
I have no idea who use directly this dependency, but should it be concidered to just drop `net46x` and `net45` ?
The same way `Major` version and pinning works, and also since `net472/net48` are supposed to be `netstandard2.0` they should show be able to resolves `2.1` dependencies


test/Serilog.Settings.Configuration.Tests/ConfigurationReaderTests.cs
```diff
@@ -14,7 +14,7 @@ public class ConfigurationReaderTests
         public ConfigurationReaderTests()
         {
             _configurationReader = new ConfigurationReader(
-                JsonStringConfigSource.LoadSection(@"{ 'Serilog': {  } }", "Serilog"),
+                JsonStringConfigSource.LoadSection(@"{ ""Serilog"": {  } }", "Serilog"),
```
sungam3r: [This](https://github.com/serilog/serilog-settings-configuration/pull/203/files?file-filters%5B%5D=.cs&file-filters%5B%5D=.md&file-filters%5B%5D=.png&file-filters%5B%5D=.ps1&file-filters%5B%5D=.yml#diff-7bebd3e3fb423271f72e33bd8524a4e4R14) may be helpful. Still waiting for merge :(

tebeco: i did not saw that PR ;)
Also the double quote seems fine for all TFM, do we want to have different JSON delimiter since one works for all ?

sungam3r: My interest was to keep the number of changes in PR to a minimum. I achieved this with a few lines of code instead of a lot of refactoring. As a result, the tests work on both platforms. In addition, you will simply stop testing a single quote syntax. This is degradation of the test coverage.

PR #212, opened 2020-03-28
New syntax for level overrides

sungam3r:
fixes https://github.com/serilog/serilog-aspnetcore/issues/184


nblumhardt:
Thanks for the PR, I think this case is too esoteric for us to build in support for right now.

sungam3r:
As you wish.

PR #218, opened 2020-05-10
NU1605 fix for .net core clients

skomis-mm:
- fixes [#196](https://github.com/serilog/serilog-settings-configuration/issues/196)

Notes: the issue occurs when `RuntimeIdentifier` specified on publish/restore for `netstandard2.x` project with `Microsoft.NETCore.Targets=1.1.x` resolved by nuget dependency resolution. This change force to use `3.0.0` as suggested in [docs](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1605#example-2) and [related](https://github.com/dotnet/runtime/issues/25664) issue.

Cosmetic adjustments:
- fixed failing tests for net452 in VS TestExplorer
- added source link support
- for .NET Framework targets added reference assemblies packages and adjusted build script to be able to build projects under Linux environment

Issue #197, opened 2019-09-11
Serilog.Settings.Configuration 3.1.0 transitivly brings Newtonsoft.Json 9.0.1


GeneralGDA:
Serilog.Settings.Configuration transitively brings (via Microsoft.Extensions.DependencyModel 2.0.4) Newtonsoft.Json 9.0.1. And it is painful to use both Serilog.Settings.Configuration and new versions of Newtonsoft.Json in a single project (it is impossible, actually). And brings constant troubles in multi-project solutions.

nblumhardt:
Thanks for the note. Some alternative dependencies/targeting that avoids this seems like it would be worthwhile 👍

PR #222, opened 2020-05-12
Update Microsoft.Extensions.DependencyModel to v3

skomis-mm:
- fixes [#197](https://github.com/serilog/serilog-settings-configuration/issues/197) to some extent for `netstandard2.0` target: the new package uses `System.Text.Json` instead of  `Newtonsoft.Json` to parse `.deps.json` files.
- no need to use workaround from [#218](https://github.com/serilog/serilog-settings-configuration/pull/218) since the new package doesn't bring .NET Core 1.x transitive dependencies, as `2.0.4` of `Microsoft.Extensions.DependencyModel` does.

### DIFF
--- a/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
+++ b/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
@@ -30,15 +30,11 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
-    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="2.0.4" />
+    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="3.0.0" />
     <PackageReference Include="Serilog" Version="2.6.0" />
     <None Include="..\..\assets\icon.png" Pack="true" PackagePath=""/>
   </ItemGroup>
-
-  <ItemGroup Condition="$(TargetFramework) == 'netstandard2.0'">
-    <PackageReference Include="Microsoft.NETCore.Targets" Version="3.0.0" />
-  </ItemGroup>
-
+  
   <ItemGroup Condition="'$(TargetFramework)' == 'net451'">
     <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="1.1.2" />
   </ItemGroup>


Ted-chiptech:
Start to have this error message in Azure Function V3:

Could not load file or assembly 'Microsoft.Extensions.DependencyModel, Version=3.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.

nblumhardt:
Hi @Ted-chiptech  - thanks for the heads-up. Can you please let us know which versions you're dealing with (packages, TFMs and runtime), along with a stack trace, in a new issue? May be a regression, but we'll need more info to pin it down. Thanks!

Ted-chiptech:
> Hi @Ted-chiptech - thanks for the heads-up. Can you please let us know which versions you're dealing with (packages, TFMs and runtime), along with a stack trace, in a new issue? May be a regression, but we'll need more info to pin it down. Thanks!

Hi @nblumhardt, I was using Serilog.Settings.Configuration V3.2.0, The Azure Function runtime is V3, TargetFramework is netcoreapp3.1.  then I will get the error above. (The error will show up immediately when starting the project)

However, if I roll back the Serilog.Settings.Configuration package to V3.1.0, everything works fine. Since the error is about 'Microsoft.Extensions.DependencyModel', I think it probably has something to do with this PR. Thanks!

skomis-mm:
Just checked, the Azure Functions host tied to v2.1.0 of `Microsoft.Extensions.DependencyModel` and the new version of `Serilog.Settings.Configuration` brings `3.0.0` to the function's `*.deps.json` which affects the host. 😕 

Not sure yet what is the best way to handle this other than to revert this change..

nblumhardt:
Nasty... Since Azure Functions will no doubt be updated someday, another option is to continue supporting it via 3.1.0, for now ....

skomis-mm:
@Ted-chiptech , @nblumhardt the [source](https://github.com/Azure/azure-functions-host/issues/5894) issue.
To summarize, there is a couple of options:
1. downgrade `Microsoft.NET.Sdk.Functions` to version `3.0.3` or
2. add undocument build property to `.csproj`
```xml
<_FunctionsSkipCleanOutput>true</_FunctionsSkipCleanOutput>
```
3.  see if you can migrate to [out-of-process](https://docs.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide) executable

thyde1:
> @Ted-chiptech , @nblumhardt the [source](https://github.com/Azure/azure-functions-host/issues/5894) issue.
> To summarize, there are a couple of options:
> 
> 1. downgrade `Microsoft.NET.Sdk.Functions` to version `3.0.3` or
> 2. add undocument build property to `.csproj`
> 
> ```
> <_FunctionsSkipCleanOutput>true</_FunctionsSkipCleanOutput>
> ```
> 
> 1. see if you can migrate to [out-of-process](https://docs.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide) executable

Alternatively you could include
```xml
<ItemGroup>
    <FunctionsPreservedDependencies Include="Microsoft.Extensions.DependencyModel.dll" />
</ItemGroup>
```
in your functions `.csproj` to preserve only _this_ dependency, and not _all_, as using `_FunctionsSkipCleanOutput` would achieve.

PR #233, opened 2020-08-28
Minimumlevel partial fix with IConfigurationRoot.Providers

HexHunter97:
A partial fix for minimumlevel not being correctly overriden by higher priority providers.

The fix only works for the .NET 4.6.1+ and .NET Standard builds as the IConfigurationRoot interface did not have Providers in .NET 4.5.1. 

I understand this is more of an annoyance than a real issue, but it's easy to miss.

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -22,12 +22,19 @@ namespace Serilog.Settings.Configuration
         readonly IConfigurationSection _section;
         readonly IReadOnlyCollection<Assembly> _configurationAssemblies;
         readonly ResolutionContext _resolutionContext;
+#if NETSTANDARD || NET461
+        readonly IConfigurationRoot _configurationRoot;
+#endif
 
         public ConfigurationReader(IConfigurationSection configSection, AssemblyFinder assemblyFinder, IConfiguration configuration = null)
         {
             _section = configSection ?? throw new ArgumentNullException(nameof(configSection));
             _configurationAssemblies = LoadConfigurationAssemblies(_section, assemblyFinder);
             _resolutionContext = new ResolutionContext(configuration);
+#if NETSTANDARD || NET461
+            _configurationRoot = configuration as IConfigurationRoot;
+#endif
+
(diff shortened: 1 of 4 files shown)


src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -22,12 +22,19 @@ class ConfigurationReader : IConfigurationReader
...
 
         public ConfigurationReader(IConfigurationSection configSection, AssemblyFinder assemblyFinder, IConfiguration configuration = null)
         {
             _section = configSection ?? throw new ArgumentNullException(nameof(configSection));
             _configurationAssemblies = LoadConfigurationAssemblies(_section, assemblyFinder);
             _resolutionContext = new ResolutionContext(configuration);
+#if NETSTANDARD || NET461
+            _configurationRoot = configuration as IConfigurationRoot;
+#endif
+
```
sungam3r, line 37:
```suggestion
```

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -124,6 +131,35 @@ void ApplyMinimumLevel(IConfigurationSection directive, Action<LoggerMinimumLeve
...
+                        }
+                    }
+
+                    return null;
+                }
+
+                #endif //NET451 or fallback
+
+                return minimumLevelDirective.Value != null ? minimumLevelDirective : minimumLevelDirective.GetSection("Default");
+
```
sungam3r, line 161:
```suggestion
```

nblumhardt:
Thanks! 👍  As it's a slightly tricky scenario, it would be great to have a test for this behavior, so we can lessen the risk of regression in the future.

XHUN97terX:
I realise the long lines are less than desirable, but I don't know how to break them up properly wihout without taking up too much vertical space. :(
Any advice on the matter welcome.

nblumhardt:
Sorry about the long delay on this; merging to the 3.5.0-dev-* version now.

HexHunter97:
It's ok, there are other priorities sometimes.
Thanks for letting it in!

Issue #239, opened 2020-10-20
AssemblyFinder throw Exception in .Net5.0(RC2) SingleFile


xljiulang:
Unhandled exception. System.NotSupportedException: CodeBase is not supported on assemblies loaded from a single-file bundle.
   at System.Reflection.RuntimeAssembly.get_CodeBase()
   at Microsoft.Extensions.DependencyModel.DependencyContextLoader.GetNormalizedCodeBasePath(Assembly assembly)
   at Microsoft.Extensions.DependencyModel.DependencyContextLoader.GetDepsJsonPath(Assembly assembly)
   at Microsoft.Extensions.DependencyModel.DependencyContextLoader.LoadAssemblyContext(Assembly assembly, IDependencyContextReader reader)
   at Microsoft.Extensions.DependencyModel.DependencyContextLoader.Load(Assembly assembly)
   at Microsoft.Extensions.DependencyModel.DependencyContext.Load(Assembly assembly)
   at Microsoft.Extensions.DependencyModel.DependencyContext.LoadDefault()
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.get_Value()
   at Microsoft.Extensions.DependencyModel.DependencyContext.get_Default()
   at Serilog.Settings.Configuration.Assemblies.AssemblyFinder.Auto()
   at Serilog.ConfigurationLoggerConfigurationExtensions.Configuration(LoggerSettingsConfiguration settingConfiguration, IConfiguration configuration, String sectionName, DependencyContext dependencyContext)
   at Serilog.ConfigurationLoggerConfigurationExtensions.Configuration(LoggerSettingsConfiguration settingConfiguration, IConfiguration configuration, DependencyContext dependencyContext)
   at HostApp.Program.<>c.<CreateHostBuilder>b__1_2(WebHostBuilderContext hosting, LoggerConfiguration logger)
   at Serilog.SerilogWebHostBuilderExtensions.<>c__DisplayClass1_0.<UseSerilog>b__0(WebHostBuilderContext context, IServiceCollection collection)
   at Microsoft.AspNetCore.Hosting.GenericWebHostBuilder.<>c__DisplayClass11_0.<ConfigureServices>b__0(HostBuilderContext context, IServiceCollection builder)
[...]

Issue #237, opened 2020-09-19
Unable to find a method called ''. Candidate methods are:


bruno-garcia:
When updating [NuGetTrends](https://github.com/dotnet/nuget-trends/tree/2da4f22ad3817772ea7573baff435de22b96fca7/src/NuGetTrends.Scheduler)'s `Serilog.Settings.Configuration` from `3.1.0` to `3.2.0-dev-00244`, one of the sinks stopped working.

The configuration is public, but basically [this sink](https://github.com/dotnet/nuget-trends/blob/2da4f22ad3817772ea7573baff435de22b96fca7/src/NuGetTrends.Scheduler/appsettings.Staging.json#L20-L28) is the culprit:

```
"Name": "LogzIo",
"Args": {
	...
}
```

From the package:

```
<PackageReference Include="Serilog.Sinks.Logz.Io" Version="2.2.1" />
```

I have two sinks there configured via appsettings.json and the other (Sentry) works properly.

<details>
  <summary>error log</summary>

```
2020-09-19T22:07:22.3101180Z Unable to find a method called LogzIo. Candidate methods are:
Serilog.LoggerConfiguration Sentry(Serilog.Configuration.LoggerSinkConfiguration, System.String, Serilog.Events.LogEventLevel, Serilog.Events.LogEventLevel, System.IFormatProvider, System.Nullable`1[System.Boolean], System.Nullable`1[System.Boolean], System.String, System.Nullable`1[System.Boolean], System.Nullable`1[System.Int32], System.Nullable`1[System.Single], System.String, System.String, System.Nullable`1[System.Int32], System.Nullable`1[System.TimeSpan], System.Nullable`1[System.Net.DecompressionMethods], System.Nullable`1[System.IO.Compression.CompressionLevel], System.Nullable`1[System.Boolean], System.Nullable`1[System.Boolean], System.Nullable`1[Sentry.Protocol.SentryLevel], System.Nullable`1[System.Boolean], System.Nullable`1[Sentry.DeduplicateMode], System.Nullable`1[System.Boolean])
Serilog.LoggerConfiguration Sentry(Serilog.Configuration.LoggerSinkConfiguration, System.Action`1[Sentry.Serilog.SentrySerilogOptions])
[...]

PR #241, opened 2020-11-17
Configuration assembly auto-discovery fixes

skomis-mm:
- fixes [#239](https://github.com/serilog/serilog-settings-configuration/issues/239), [#237](https://github.com/serilog/serilog-settings-configuration/issues/237)
- added .net 5.0 tfm
- added `Serilog.Expressions` to the sample project
- documentation update

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 * #219 - reduced search graph for configuration dlls to avoid native assets
 * #221 - added support for conditional/leveled enrichers from Serilog 2.9+
 * #222 - updated Microsoft.Extensions.DependencyModel
+* #237 - DependencyContextAssemblyFinder fix: check `serilog` in any part of the name for any dependent package
+* #239 - handle NotSupportedException for .net 5.0 single file applications
 
 3.1.0
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@
 * #219 - reduced search graph for configuration dlls to avoid native assets
 * #221 - added support for conditional/leveled enrichers from Serilog 2.9+
 * #222 - updated Microsoft.Extensions.DependencyModel
-* #237 - DependencyContextAssemblyFinder fix: check `serilog` in any part of the name for any dependent package
+* #237 - DependencyContextAssemblyFinder fix: check `serilog` at the start of the name for any dependent package
 * #239 - handle NotSupportedException for .net 5.0 single file applications
 
 3.1.0
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/Assemblies/AssemblyFinder.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/Assemblies/AssemblyFinder.cs
@@ -16,12 +16,19 @@ namespace Serilog.Settings.Configuration.Assemblies
 
         public static AssemblyFinder Auto()
         {
-            // Need to check `Assembly.GetEntryAssembly()` first because 
-            // `DependencyContext.Default` throws an exception when `Assembly.GetEntryAssembly()` returns null
-            if (Assembly.GetEntryAssembly() != null && DependencyContext.Default != null)
+            try
             {
-                return new DependencyContextAssemblyFinder(DependencyContext.Default);
+                // Need to check `Assembly.GetEntryAssembly()` first because 
+                // `DependencyContext.Default` throws an exception when `Assembly.GetEntryAssembly()` returns null
+                if (Assembly.GetEntryAssembly() != null && DependencyContext.Default != null)
+                {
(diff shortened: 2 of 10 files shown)


src/Serilog.Settings.Configuration/Settings/Configuration/Assemblies/DependencyContextAssemblyFinder.cs
```diff
@@ -27,7 +27,7 @@ where IsCaseInsensitiveMatch(assemblyName.Name, nameToFind)
             
             static bool IsReferencingSerilog(Library library)
             {
-                return library.Dependencies.Any(dependency => dependency.Name.Equals("serilog", StringComparison.OrdinalIgnoreCase));
+                return library.Dependencies.Any(dependency => dependency.Name.IndexOf("serilog", StringComparison.OrdinalIgnoreCase) >= 0);
```
tsimbalar, line 30: not 100% sure about the impact here 🤔 

Could it be problematic if there is a package like `SomeRandomThingWithTheNameSerilogInIt` ? should we stick to "starts with" serilog maybe ?

skomis-mm: makes sense 👍

src/Serilog.Settings.Configuration/Settings/Configuration/Assemblies/AssemblyFinder.cs
```diff
@@ -16,12 +16,19 @@ protected static bool IsCaseInsensitiveMatch(string text, string textToFind)
...
+                // `DependencyContext.Default` throws an exception when `Assembly.GetEntryAssembly()` returns null
+                if (Assembly.GetEntryAssembly() != null && DependencyContext.Default != null)
+                {
+                    return new DependencyContextAssemblyFinder(DependencyContext.Default);
+                }
             }
+            catch (NotSupportedException) when (typeof(object).Assembly.Location is "") // bundled mode detection
+            {
+            }
+
```
tsimbalar, line 31: should this actually throw instead of ignoring silently ? or writing a message to SelfLog or something with a hint about what needs to be done ?

src/Serilog.Settings.Configuration/Settings/Configuration/Assemblies/DependencyContextAssemblyFinder.cs
```diff
@@ -27,7 +27,10 @@ where IsCaseInsensitiveMatch(assemblyName.Name, nameToFind)
             
             static bool IsReferencingSerilog(Library library)
             {
-                return library.Dependencies.Any(dependency => dependency.Name.IndexOf("serilog", StringComparison.OrdinalIgnoreCase) >= 0);
+                const string Serilog = "serilog";
+                return library.Dependencies.Any(dependency =>
+                    dependency.Name.StartsWith(Serilog, StringComparison.OrdinalIgnoreCase) &&
+                   (dependency.Name.Length == Serilog.Length || dependency.Name[Serilog.Length] == '.'));
```
tsimbalar, line 33: I think I can see some inspiration from 

https://github.com/serilog/serilog/blob/e82213e98ece04bdfcb13d066f91b987ca4921c6/src/Serilog/Core/LevelOverrideMap.cs#L76-L78
🤣

src/Serilog.Settings.Configuration/Settings/Configuration/Assemblies/AssemblyFinder.cs
skomis-mm: It throws currently :) AFAIK there is no good workaround for single file deployments, except using `Using` section. I've updated the docs.

The problem with writing to `SelfLog` is it will show message with correct configuration (because of using `Using` section or not using external assemblies).

tsimbalar: oh right, nevermind then !

tsimbalar (approved):
LGTM ! :shipit:

CHANGES.md
```diff
@@ -8,6 +8,8 @@
 * #219 - reduced search graph for configuration dlls to avoid native assets
 * #221 - added support for conditional/leveled enrichers from Serilog 2.9+
 * #222 - updated Microsoft.Extensions.DependencyModel
+* #237 - DependencyContextAssemblyFinder fix: check `serilog` in any part of the name for any dependent package
```
sungam3r, line 11: Not true now:
> return library.Dependencies.Any(dependency =>
                    dependency.Name.StartsWith(Serilog, StringComparison.OrdinalIgnoreCase) &&
                   (dependency.Name.Length == Serilog.Length || dependency.Name[Serilog.Length] == '.'));

PR #242, opened 2020-11-23
make '$' sign optional for minimum level / filter switch declarations

### LINKED ISSUE #231: Unable to override log level switch initial log level via environment variables
Since we deploy our services in k8s we need to override level switch initial log level via environment variable. At the same time levelSwitchName must start with `$` sign which is not allowed for env variable name.

How can initialLevel can be overriden?

Example:
1) app config:
```json
  "Serilog": {
    "levelSwitches": {
      "$logLevelSwitch": "Debug"
    },
    "MinimumLevel": {
      "controlledBy": "$logLevelSwitch"
...
```
2) override:
```yaml
env:
  name: serilog_levelSwitches_$logLevelSwitch
  value: Debug
```

Error:
`"Invalid value: "serilog__levelSwitches__$logLevelSwitch": a valid environment variable name must consist of alphabetic characters, digits, '_', '-', or '.', and must not start with a digit"`

skomis-mm:
- fixes #231 , implements both [Option 2 and 3](https://github.com/serilog/serilog-settings-appsettings/issues/7#issuecomment-325889701) of the initial design
- update reference to Serilog 2.10.0
- documentation update

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,12 +2,14 @@
 
 3.2.0 (pre-release)
 
+* #162 - LoggingFilterSwitch support
 * #202 - added support to AuditTo.Logger
 * #203 - added support for custom types in arrays and custom collections
 * #218 - fixed an issue with `dotnet restore` with `rid` specified if referenced from `netstandard` project
 * #219 - reduced search graph for configuration dlls to avoid native assets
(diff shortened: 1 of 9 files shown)


tsimbalar (approved):
LGTM 💪

PR #271, opened 2021-07-06
Added few examples

tnsatish:


nblumhardt:
Thanks for sending these! I agree that the README needs some love, but I don't think given the very wide variety of possible configurations and the verbose syntax, we're best off leaving further examples to Stack Overflow, unless we can come up with a dense/structured way of representing them. Thanks again!

PR #311, opened 2022-05-09
Remove out of support target frameworks

0xced:
* Add .NET 6 target framework instead
* Update tests and sample dependencies
* Simplify conditional package references in all projects
* Remove the custom `PRIVATE_BIN` define, use the built-in `NETFRAMEWORK` instead
* Also use the Visual Studio 2022 image on AppVeyor because VS 2022 is required for the .NET 6 SDK
* Add a reference to `Microsoft.TestPlatform.ObjectModel` so that `dotnet test` also works on Linux and macOS for the .NET Framework targets
* Depend on Microsoft.Extensions.Configuration.Binder: there's no need to depend on Microsoft.Extensions.Options.ConfigurationExtensions, taking a dependency on Microsoft.Extensions.Configuration.Binder is enough for what Serilog.Settings.Configuration needs to do.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -6,6 +6,6 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.2" PrivateAssets="all" />
   </ItemGroup>
 </Project>
--- a/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
+++ b/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
@@ -24,22 +24,18 @@
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
-  <PropertyGroup Condition="('$(TargetFramework)' == 'net451') Or ('$(TargetFramework)' == 'net461')">
-    <DefineConstants>$(DefineConstants);PRIVATE_BIN</DefineConstants>
-  </PropertyGroup>
-
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
     <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="3.0.0" />
     <PackageReference Include="Serilog" Version="2.10.0" />
-    <None Include="..\..\assets\icon.png" Pack="true" PackagePath=""/>
-  </ItemGroup>
-  
-  <ItemGroup Condition="'$(TargetFramework)' == 'net451'">
-    <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="1.1.2" />
+    <None Include="..\..\assets\icon.png" Pack="true" PackagePath="" />
   </ItemGroup>
 
-  <ItemGroup Condition="('$(TargetFramework)' == 'netstandard2.0') Or ('$(TargetFramework)' == 'net461')">
-    <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="2.0.0" />
+  <ItemGroup Condition="$(TargetFramework) == 'net451'">
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="1.1.2" />
+  </ItemGroup>
+
+  <ItemGroup Condition="$(TargetFramework) != 'net451'">
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="2.0.0" />
   </ItemGroup>
 </Project>
--- a/Build.ps1
+++ b/Build.ps1
@@ -43,9 +43,8 @@ foreach ($test in dir test/*.Tests) {
     echo "build: Testing project in $test"
 
     if ($PSVersionTable.Platform -eq "Unix") {
-        & dotnet test -c Release -f netcoreapp2.1
         & dotnet test -c Release -f netcoreapp3.1
-        & dotnet test -c Release -f net50
+        & dotnet test -c Release -f net6.0
     } else {
         & dotnet test -c Release
     }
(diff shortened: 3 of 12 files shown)


src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
```diff
@@ -5,7 +5,7 @@
     <VersionPrefix>3.3.1</VersionPrefix>
```
nblumhardt, line 5: This change would require a major version bump, e.g. to 4.0.0

src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
```diff
@@ -5,7 +5,7 @@
...
     <LangVersion>latest</LangVersion>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>netstandard2.0;net451;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net462</TargetFrameworks>
```
nblumhardt, line 8: Since 6.0 is LTS and brings in fewer dependencies, it might make sense to target it specifically here.

nblumhardt (commented):
I'm okay with this change if it makes it easier to maintain this package going forwards. Any thoughts @skomis-mm ?

src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
0xced: You mean, because of dropping the `net451` and `net461` target frameworks? I have read both [.NET fundamentals ⟩ Library guidance ⟩ Breaking changes](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/breaking-changes) and [.NET fundamentals ⟩ Changes that affect compatibility](https://docs.microsoft.com/en-us/dotnet/core/compatibility/) which give very good advice but none of them mentions dropping a target framework.

src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
0xced: > brings in fewer dependencies

What do you mean? After this pull request there are only 3 dependencies, all included unconditionally: `Microsoft.Extensions.Configuration.Binder`, `Microsoft.Extensions.DependencyModel` and `Serilog`.

I see many package authors add `net6.0` to the target frameworks and I don't understand why. I see why you'd target .NET 6 if you are making use of new features introduced in .NET 6 such as the new `DateOnly` and `TimeOnly` types. If there's no `#if NET6_0_OR_GREATER` in the code, I don't see the point of targeting .NET 6, but maybe I'm missing something and I'd be happy to be enlightened.

nblumhardt: You might be right; I'll need to dig back into it :-)

Last time I checked, I think the NuGet dependency resolution algorithm would mean that a package targeting netstandard2.0 might drag in more transitive dependencies than a package targeting the later TFM, but it's a hazy memory and might not matter here.

Let's leave it out, for now 👍

0xced: I think what you are referring to was the situation before .NET Standard 2.0 where many `System.*` packages were brought in. This issue has been solved with .NET Standard 2.0 so indeed it doesn't matter here. There's a good discussion about that on Stack Overflow: [Why does my .NET Standard NuGet package trigger so many dependencies?](https://stackoverflow.com/questions/47365136/why-does-my-net-standard-nuget-package-trigger-so-many-dependencies/47366401#47366401)

nblumhardt: Thanks for the pointer; no, I remember the switch away from NETStandard.Library, my memory is of someone coming along after that and demonstrating to me why netstandard2.0 still wasn't the optimal TFM ... 🤔 I'll try to find that thread when I have some time :-)

Recently I've been enabling 5 or 6 when turning on non-null reference type checking, which avoids bundling types from the Nullable package on platforms that already carry them. Not relevant to the change here, though.

src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
nblumhardt: Dropping a framework will mean that users on .NET 4.5 (for example) won't be able to install the package, so it'll definitely be a breaking change from a SemVer perspective. I think because it's a big/deliberate breakage it's not really in the scope of what those docs are discussing. Let me know if I'm missing something here, though.

0xced: Yeah, you're totally right. I just tried to add those properties in the `Serilog.Settings.Configuration.csproj` file in order to perform [package validation](https://docs.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview):

```xml
<PropertyGroup>
  <EnablePackageValidation>true</EnablePackageValidation>
  <PackageValidationBaselineVersion>3.3.0</PackageValidationBaselineVersion>
</PropertyGroup>
```

Indeed, this is considered a breaking change. Here's the error after running `dotnet pack`:
> /usr/local/share/dotnet/sdk/6.0.202/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Compatibility.Common.targets(32,5): error PKV006: Target framework .NETFramework,Version=v4.5.1 is no longer supported in the latest version. [serilog-settings-configuration/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj]

skomis-mm: I think unless as it doesn't cost anything from maintanence perspective point of view we can still support older TFM's even if they are eol. Bumping to 4.0 ideally would be good to be done with having some new features first, so it would not took long time to deliver stable release.

0xced: I rebased on the `dev` branch, did the EOL target framework cleanup but only for the tests and sample projects in order not to introduce a breaking change.

Directory.Build.props
```diff
@@ -4,8 +4,4 @@
     <TreatWarningsAsErrors>True</TreatWarningsAsErrors>
     <TreatSpecificWarningsAsErrors />
   </PropertyGroup>
-
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="all" />
-  </ItemGroup>
 </Project>
```
skomis-mm, line 7: Probably we can leave this? Not sure why, but VS on Windows still complaints about missing targeting packs:
```csharp
C:\Program Files\Microsoft Visual Studio\2022\Preview\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets(1221,5): error MSB3644: The reference assemblies for .NETFramework,Version=v4.6.2 were not found
```

0xced: Argh! I have amended that in 3f9593e1c70a37b57f9c5e01ef1284b12cf8cda9. I use Rider and not Visual Studio, that's why I missed this issue which is common apparently: https://github.com/dotnet/runtime/issues/68716

src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
0xced: Interesting post about  .NET Standard 2.0 vs .NET 6 targeting on Stack Overflow for the record: [Are there benefits in producing a .NET 6.0 version of a .NET Standard 2.0 library?](https://stackoverflow.com/questions/70778797/are-there-benefits-in-producing-a-net-6-0-version-of-a-net-standard-2-0-librar/72266562#72266562)

PR #339, opened 2022-11-10
Using different versions of Microsoft.Extensions.DependencyModel base…

hypdeb:
Using different versions of `Microsoft.Extensions.DependencyModel` based on target framework. This resolves warnings when using package as part of solution where another package requires a more recent version of `Microsoft.Extensions.DependencyModel`.

### DIFF
--- a/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
+++ b/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
@@ -26,7 +26,6 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="3.0.0" />
     <PackageReference Include="Serilog" Version="2.10.0" />
     <None Include="..\..\assets\icon.png" Pack="true" PackagePath="" />
   </ItemGroup>
@@ -38,4 +37,17 @@
   <ItemGroup Condition="$(TargetFramework) != 'net451'">
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="2.0.0" />
   </ItemGroup>
+
+  <ItemGroup Condition="$(TargetFramework) == 'net451' Or $(TargetFramework) == 'net461'">
+    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="3.0.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="$(TargetFramework) == 'net7.0'">
+    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="7.0.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="$(TargetFramework) != 'net451' And $(TargetFramework) != 'net461' And $(TargetFramework) != 'net7.0'">
+    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="6.0.0" />
+  </ItemGroup>
+
 </Project>
--- a/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
+++ b/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
@@ -5,7 +5,7 @@
     <VersionPrefix>3.5.0</VersionPrefix>
     <LangVersion>latest</LangVersion>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>netstandard2.0;net451;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net451;net461;net7.0</TargetFrameworks>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AssemblyName>Serilog.Settings.Configuration</AssemblyName>
--- a/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
+++ b/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
@@ -5,7 +5,7 @@
     <VersionPrefix>3.5.0</VersionPrefix>
     <LangVersion>latest</LangVersion>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>netstandard2.0;net451;net461;net7.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net451;net461</TargetFrameworks>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AssemblyName>Serilog.Settings.Configuration</AssemblyName>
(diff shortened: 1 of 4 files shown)


src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
```diff
@@ -38,4 +37,17 @@
   <ItemGroup Condition="$(TargetFramework) != 'net451'">
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="2.0.0" />
   </ItemGroup>
+
+  <ItemGroup Condition="$(TargetFramework) == 'net451' Or $(TargetFramework) == 'net461'">
+    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="3.0.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="$(TargetFramework) == 'net7.0'">
```
sungam3r, line 45: You forgot to add net7.0 into `TargetFrameworks`

hypdeb: I added the `net7.0` target framework. One test seems to fail and I would need either time or help with it. I don't understand what's going on so far. It seems the `MixedMinimumLevel` is not resolved correctly in this case.

sungam3r: изображение

SimonCropp:
IMO we should just update to Microsoft.Extensions.DependencyModel v7. with no conditions. if people want to use older versions they can use older nugets

nblumhardt:
I'm coming around to that point of view too, @SimonCropp.

To do that well, we'd need to synchronize major version numbers here with those in Serilog.Extensions.Logging, .Hosting, and Serilog.AspNetCore - i.e. make it consistent across all of the packages tied to the app model. Can you think of anything that's likely to spoil this kind of plan?

sungam3r:
> if people want to use older versions they can use older nugets

This is a wrong interpretation. People do not want to use old versions, sometimes they are forced to use them due to various restrictions. It’s simply pointless to increase the major version of dependency without a good reason.

SimonCropp:
how will this work if someone has to use Microsoft.Extensions.DependencyModel v3 on net7

src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
hypdeb: There doesn't seem to be a version of the visual studio image supporting `.net7.0`. The latest I see there only supports `6.0.402`.

hypdeb:
Anyways, it doesn't seem possible to support `.net7.0` directly currently due to `AppVeyor`'s Visual Studio images not supporting it. I reverted those changes.

SimonCropp:
also how will this work if someone has to use Microsoft.Extensions.DependencyModel v7 on net4.6.2

hypdeb:
> also how will this work if someone has to use Microsoft.Extensions.DependencyModel v7 on net4.6.2

The same thing that is happening now for all framework versions: you get a warning that there's a conflict that can't be resolved between 3.0.0 required by this library and 7.0.0 required by you or some other library.

Moreover, trying to use Microsoft.Extensions.DependencyInjection 7.0.0 in a net462 app will produce another warning that the library doesn't support this target framework / has not been tested with it.

kavun:
@hypdeb, you should be able to add `net7.0` back now 🥳 https://www.appveyor.com/updates/
image

hypdeb:
@kavun thanks, sorry for the delay to apply this change. But I get one failing test locally with .net 6 and .net 7. I will try to have a look tonight.

glen-84:
Will this be merged? Seeing a wall of warnings during builds because of this.

hypdeb:
Sorry I kind of forgot about this because the tooling used to build this project did not support the latest version of .NET. I'll try to have another look now.

hypdeb:
Sadly some tests still fail. It seems some non-backwards compatible change was made in `DependencyModel`.
image

glen-84:
Seems like only one test – `MixedMinimumLevelCorrectOneIsEnabledOnLogger`.

glen-84:
This is actually a recently added test by @HexHunter97. Maybe they can help.

0xced:
The error in the `MixedMinimumLevelCorrectOneIsEnabledOnLogger` test is unrelated to the `Microsoft.Extensions.DependencyModel` package version.

Only adding `net6.0` in the  `TargetFrameworks` of `Serilog.Settings.Configuration.csproj` is enough to produces this error.

I don't understand why (yet)!

hypdeb:
@0xced Thanks for your help! I was already digging in the wrong direction.

HexHunter97:
I'll try take a look in a few hours once free from work to confirm, but my suspicion is this:

It probably has something to do with the preprocessor directives in the ConfigurationReader and ConfigurationReaderTests not matching. 
At the time I wrote this piece of code in 2020, there was only NETSTANDARD, NET452 and NET461 versions of the code and didn't think of this potential conflict and didn't know about MS's plans to fully switch to .NET.

https://github.com/serilog/serilog-settings-configuration/blob/ab8764198c7fc784f1a02f3c2d02e53532204941/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs#L23

https://github.com/serilog/serilog-settings-configuration/blob/ab8764198c7fc784f1a02f3c2d02e53532204941/test/Serilog.Settings.Configuration.Tests/ConfigurationReaderTests.cs#L225

A fix would be to add NET as the preprocessor directive in ConfigurationReader.

So as said, I'll confirm this in a few hours.

0xced:
Very good point!

I always try to make conditional compilation future proof. See https://github.com/spectreconsole/spectre.console/pull/563 for an example how to achieve this goal. The same principles should be applied in `Serilog.Settings.Configuration` too.

0xced:
Yep, that's it! Replacing `#if NETSTANDARD || NET461` with `#if !NET451` everywhere fixes the failing test when adding the .NET 6 target framework.

HexHunter97:
This one's on me completely. But you live & learn. 😄 
Apologies for the headaches guys!

HexHunter97:
This one's on me completely. But you live & learn. 😄 
Apologies for the headaches guys!

hypdeb:
Indeed, you could also add `|| NET6_OR_GREATER`

hypdeb:
But @0xced 's solution is more explicit.

hypdeb:
All green now! Can someone give it a final review and merge it?

0xced:
OK now I see that Microsoft.Extensions.Configuration.Binder 2.0.0 and Microsoft.Extensions.DependencyModel 3.0.0 have been deprecated as part of the [.NET Package Deprecation effort](https://github.com/dotnet/announcements/issues/217).

So I'd be in favour of simply dropping the `net451` target framework, bumping both Microsoft.Extensions.Configuration.Binder and Microsoft.Extensions.DependencyModel to version 6.0.0 and removing all conditional compilation.

I just proposed this alternative solution in #351 which should supersede this pull request in my humble opinion.

hypdeb:
@0xced Yes please, I didn't know this was an option!

glen-84:
@0xced

For us it started happening when we moved our tests to a separate project, and that test project referenced the main application project.

image

I need to check the list of dependencies again.

PR #347, opened 2023-02-06
Add a way to get the configured log level switches

### LINKED ISSUE #206: How to configure from JSON, but then get a reference to the switch?
I'd like to use the json configuration to configure the seq sink. However I need access to the switch in code, because I want to allow this to be changed at runtime in code. How should I go about that?

0xced:
I have created this pull request as a draft because a few things should be discussed before merging.

1. I chose `LoadedConfiguration` for the class holding the loaded log level switches, as proposed by @tsimbalar in https://github.com/serilog/serilog-settings-configuration/issues/206#issuecomment-643762537 I'm not convinced it's the best name and I'm happy to change it to something better.
2. This `LoadedConfiguration` class could also expose the log filter switches, although not the internal `LoggingFilterSwitchProxy` class directly. Maybe through a public `ILoggingFilterSwitch` interface with a single `Expression { get; set; }` property?

I am looking forward to your feedback.

Fixes #206

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/LoadedConfiguration.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/LoadedConfiguration.cs
@@ -15,6 +15,6 @@ public class LoadedConfiguration
     /// <summary>
     /// The log level switches that were loaded from the <c>LevelSwitches</c> section of the configuration.
     /// </summary>
-    /// <remarks>The keys of the dictionary are the name of the switches, including the leading <c>$</c> character.</remarks>
+    /// <remarks>The keys of the dictionary are the names of the switches, including the leading <c>$</c> character.</remarks>
     public IReadOnlyDictionary<string, LoggingLevelSwitch> LogLevelSwitches { get; }
 }
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/LoadedConfiguration.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/LoadedConfiguration.cs
@@ -7,16 +7,14 @@ namespace Serilog.Settings.Configuration;
 /// </summary>
 public class LoadedConfiguration
 {
-    readonly IDictionary<string, LoggingLevelSwitch> _logLevelSwitches;
-
-    internal LoadedConfiguration(IDictionary<string, LoggingLevelSwitch> logLevelSwitches)
+    internal LoadedConfiguration(IReadOnlyDictionary<string, LoggingLevelSwitch> logLevelSwitches)
     {
-        _logLevelSwitches = logLevelSwitches;
+        LogLevelSwitches = logLevelSwitches;
     }
 
     /// <summary>
     /// The log level switches that were loaded from the <c>LevelSwitches</c> section of the configuration.
     /// </summary>
-    /// <remarks>The keys of the dictionary are the name of the switches without the leading <c>$</c> character.</remarks>
-    public IReadOnlyDictionary<string, LoggingLevelSwitch> LogLevelSwitches => _logLevelSwitches.ToDictionary(e => e.Key.TrimStart('$'), e => e.Value);
+    /// <remarks>The keys of the dictionary are the name of the switches, including the leading <c>$</c> character.</remarks>
+    public IReadOnlyDictionary<string, LoggingLevelSwitch> LogLevelSwitches { get; }
 }
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/LoadedConfiguration.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/LoadedConfiguration.cs
@@ -1,0 +1,22 @@
+using Serilog.Core;
+
+namespace Serilog.Settings.Configuration;
+
+/// <summary>
+/// Contains information about the loaded configuration.
+/// </summary>
+public class LoadedConfiguration
+{
+    readonly IDictionary<string,LoggingLevelSwitch> _logLevelSwitches;
+
+    internal LoadedConfiguration(IDictionary<string, LoggingLevelSwitch> logLevelSwitches)
+    {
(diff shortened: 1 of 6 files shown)


src/Serilog.Settings.Configuration/Settings/Configuration/LoadedConfiguration.cs
```diff
@@ -0,0 +1,22 @@
+using Serilog.Core;
+
+namespace Serilog.Settings.Configuration;
+
+/// <summary>
+/// Contains information about the loaded configuration.
+/// </summary>
+public class LoadedConfiguration
+{
+    readonly IDictionary<string,LoggingLevelSwitch> _logLevelSwitches;
```
sungam3r, line 10:
```suggestion
    readonly IDictionary<string, LoggingLevelSwitch> _logLevelSwitches;
```

src/Serilog.Settings.Configuration/Settings/Configuration/LoadedConfiguration.cs
```diff
@@ -0,0 +1,22 @@
...
+    internal LoadedConfiguration(IDictionary<string, LoggingLevelSwitch> logLevelSwitches)
+    {
+        _logLevelSwitches = logLevelSwitches;
+    }
+
+    /// <summary>
+    /// The log level switches that were loaded from the <c>LevelSwitches</c> section of the configuration.
+    /// </summary>
+    /// <remarks>The keys of the dictionary are the name of the switches without the leading <c>$</c> character.</remarks>
+    public IReadOnlyDictionary<string, LoggingLevelSwitch> LogLevelSwitches => _logLevelSwitches.ToDictionary(e => e.Key.TrimStart('$'), e => e.Value);
```
sungam3r, line 21: Property allocates `Dictionary` and N strings each call to it.

sungam3r: I don't know if it is worth worrying about the initial dollar sign in the name. In fact, this is part of the switch name. What is the goal in removal of the dollar sign?

0xced: > In fact, this is part of the switch name.

Well, this is debatable: the `LevelSwitches` section does not need to contain the `$` prefix. As found in the README and the `appsettings.json` file of the `Sample` project.

```json
{
  "Serilog": {
    "Using": [ "Serilog.Sinks.Console" ],
    "LevelSwitches": { "controlSwitch": "Verbose" },
```

> Property allocates `Dictionary` and N strings each call to it.

I wouldn't worry too much about that, one call is made per call to `.ReadFrom.Configuration(configuration)`. So for most apps that would be exactly once at startup I guess.

Also, `IDictionary` does not implement `IReadOnlyDictionary`, _[sigh](https://github.com/dotnet/runtime/issues/31001)_. But we could store `Dictionary<string, LoggingLevelSwitch>` instead of `IDictionary<string, LoggingLevelSwitch>` everywhere if the consensus is that this allocation must really be avoided.

sungam3r: > the LevelSwitches section does not need to contain the $ prefix.

Then it makes no sense in cutting off the dollar sign if it is not a special symbol. Why to do this? Just return `_logLevelSwitches` as is.

0xced: I have addressed this concern in 2ecbe0a32c12c9d75941afcd0a6f47fdfd9bf534.

src/Serilog.Settings.Configuration/Settings/Configuration/LoadedConfiguration.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    internal LoadedConfiguration(IReadOnlyDictionary<string, LoggingLevelSwitch> logLevelSwitches)
+    {
+        LogLevelSwitches = logLevelSwitches;
+    }
+
+    /// <summary>
+    /// The log level switches that were loaded from the <c>LevelSwitches</c> section of the configuration.
+    /// </summary>
+    /// <remarks>The keys of the dictionary are the name of the switches, including the leading <c>$</c> character.</remarks>
```
sungam3r, line 18:
```suggestion
    /// <remarks>The keys of the dictionary are the names of the switches, including the leading <c>$</c> character if any.</remarks>
```

src/Serilog.Settings.Configuration/Settings/Configuration/LoadedConfiguration.cs
sungam3r: 👍  $ should be appended in tests.

0xced: Of course! Fixed in 3f459e1b353a92b92a36064ac04aa0b8d4dad945.

src/Serilog.Settings.Configuration/Settings/Configuration/LoadedConfiguration.cs
0xced: Fixed in c58ef69e085c974e22621d233133543b2d30351a, but not the _if any_ part because the switch __always__ have a leading `$` in the `_declaredLevelSwitches` dictionary:

```csharp
public void AddLevelSwitch(string levelSwitchName, LoggingLevelSwitch levelSwitch)
{
    _declaredLevelSwitches[ToSwitchReference(levelSwitchName)] = levelSwitch;
}

string ToSwitchReference(string switchName)
{
    return switchName.StartsWith("$") ? switchName : $"${switchName}";
}
```

sungam3r: Ok. I didn't know.

PR #348, opened 2023-02-09
Add support for IFormatProvider used to convert string to other types

### LINKED ISSUE #325: Configuration with decimal-point numbers
When reading configuration for sinks that include decimal-point numbers, the configuration reads the value as a string and then parses it using whatever number format the current culture supports.

Example of configuring Sentry as a sink, where SampleRate should be set to 10%::
``` "WriteTo": [
      {
        "Name": "Sentry",
        "Args": {
          "Dsn": "https://<some value>.ingest.sentry.io/<some value>",
          "MinimumEventLevel": "Warning",
          "Environment": "Prod",
          "Release": "<some value>",
          "SampleRate": 0.1
        }
      }
```
The value is read / parsed as a string [here](https://github.com/serilog/serilog-settings-configuration/blob/691aefa314c575393fa09923f7f88dbdee98f7cb/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs#L319):
`                argumentValue = new StringArgumentValue(argumentSection.Value);`

While applying the configuration the rSampleRate value then depends on whether the current context supports point or comma as decimal point, giving either 0.1 (10%) or 1 (100%) as result.

The desired behavior would be for the number value to be parsed as a number and not a string, when the value is supplied as a number.

Package versions used:

- Serilog: 2.10.0
- Serilog.Settings.Configuration: 3.3.0
- Sentry.Serilog: 3.17.1

0xced:
Also introduce a new `ConfigurationContext` class to avoid `ReadFrom.Configuration()` methods exponential growth when adding new options.

All _older_ `Configuration()` methods go through the newly introduced `Configuration(LoggerSettingsConfiguration, IConfiguration, ConfigurationContext)` method that takes an `ConfigurationContext` instance.

Older methods explicitly set the `FormatProvider` option to `null` in order to preserve backward compatibility.

By using the new `Configuration()` method, users opt into the new default of having the invariant culture as the format provider.

Note: the `= null` default value in the `Configuration()` method taking a `DependencyContext` has been removed in order to make sure the CS0121 compilation does not occur:

> [CS0121] The call is ambiguous between the following methods or properties: 'ConfigurationLoggerConfigurationExtensions.Configuration(LoggerSettingsConfiguration, IConfiguration, DependencyContext)' and 'ConfigurationLoggerConfigurationExtensions.Configuration(LoggerSettingsConfiguration, IConfiguration, ConfigurationContext)'

Fixes #325

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,9 @@ Root section name can be changed:
 ```
 
 ```csharp
+var options = new ConfigurationReaderOptions { SectionName = "CustomSection" };
 var logger = new LoggerConfiguration()
-    .ReadFrom.Configuration(configuration, sectionName: "CustomSection")
+    .ReadFrom.Configuration(configuration, options)
     .CreateLogger();
 ```
 
@@ -106,8 +107,9 @@ In case of [non-standard](#azure-functions-v2-v3) dependency management you can 
 ```csharp
 var functionDependencyContext = DependencyContext.Load(typeof(Startup).Assembly);
 
+var options = new ConfigurationReaderOptions(functionDependencyContext) { SectionName = "AzureFunctionsJobHost:Serilog" };
 var logger = new LoggerConfiguration()
-    .ReadFrom.Configuration(hostConfig, sectionName: "AzureFunctionsJobHost:Serilog", dependencyContext: functionDependencyContext)
+    .ReadFrom.Configuration(hostConfig, options)
     .CreateLogger();
 ```
 
@@ -119,8 +121,9 @@ var configurationAssemblies = new[]
     typeof(ConsoleLoggerConfigurationExtensions).Assembly,
     typeof(FileLoggerConfigurationExtensions).Assembly,
 };
+var options = new ConfigurationReaderOptions(configurationAssemblies);
 var logger = new LoggerConfiguration()
-    .ReadFrom.Configuration(configuration, configurationAssemblies)
+    .ReadFrom.Configuration(configuration, options)
     .CreateLogger();
 ```
 
@@ -282,6 +285,8 @@ Some Serilog packages require a reference to a logger configuration object. The 
 
 When the configuration specifies a discrete value for a parameter (such as a string literal), the package will attempt to convert that value to the target method's declared CLR type of the parameter. Additional explicit handling is provided for parsing strings to `Uri`, `TimeSpan`, `enum`, arrays and custom collections.
 
+Since version 4.0.0, conversion will use the invariant culture (`CultureInfo.InvariantCulture`) as long as the `ReadFrom.Configuration(IConfiguration configuration, ConfigurationReaderOptions options)` method is used. Obsolete methods use the current culture to preserve backward compatibility.
+
 ### Static member support
 
 Static member access can be used for passing to the configuration argument via [special](https://github.com/serilog/serilog-settings-configuration/blob/dev/test/Serilog.Settings.Configuration.Tests/StringArgumentValueTests.cs#L35) syntax:
@@ -377,8 +382,9 @@ public class Startup : FunctionsStartup
             var functionDependencyContext = DependencyContext.Load(typeof(Startup).Assembly);
 
             var hostConfig = sp.GetRequiredService<IConfiguration>();
+            var options = new ConfigurationReaderOptions(functionDependencyContext) { SectionName = "AzureFunctionsJobHost:Serilog" };
             var logger = new LoggerConfiguration()
-                .ReadFrom.Configuration(hostConfig, sectionName: "AzureFunctionsJobHost:Serilog", dependencyContext: functionDependencyContext)
+                .ReadFrom.Configuration(hostConfig, options)
                 .CreateLogger();
 
             return new SerilogLoggerProvider(logger, dispose: true);
--- a/src/Serilog.Settings.Configuration/ConfigurationLoggerConfigurationExtensions.cs
+++ b/src/Serilog.Settings.Configuration/ConfigurationLoggerConfigurationExtensions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2013-2016 Serilog Contributors
+// Copyright 2013-2016 Serilog Contributors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -42,6 +42,7 @@ public static class ConfigurationLoggerConfigurationExtensions
     /// <param name="dependencyContext">The dependency context from which sink/enricher packages can be located. If not supplied, the platform
     /// default will be used.</param>
     /// <returns>An object allowing configuration to continue.</returns>
+    [Obsolete("Use ReadFrom.Configuration(IConfiguration configuration, ConfigurationReaderOptions readerOptions) instead.")]
     public static LoggerConfiguration Configuration(
         this LoggerSettingsConfiguration settingConfiguration,
         IConfiguration configuration,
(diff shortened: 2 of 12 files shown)


src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationContext.cs
```diff
@@ -0,0 +1,62 @@
+using System.Globalization;
+using System.Reflection;
+using Microsoft.Extensions.DependencyModel;
+
+namespace Serilog.Settings.Configuration;
+
+/// <summary>
+/// A context to adjust how the configuration object is processed.
+/// </summary>
+public class ConfigurationContext
```
sungam3r, line 10:
```suggestion
public sealed class ConfigurationContext
```

At least for the first time.

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationContext.cs
```diff
@@ -0,0 +1,62 @@
...
+    public string SectionName { get; init; } = ConfigurationLoggerConfigurationExtensions.DefaultSectionName;
+
+    /// <summary>
+    /// The <see cref="IFormatProvider"/> used when converting strings to other object types. Defaults to the invariant culture.
+    /// </summary>
+    public IFormatProvider FormatProvider { get; init; } = CultureInfo.InvariantCulture;
+
+    internal Assembly[] Assemblies { get; }
+    internal DependencyContext DependencyContext { get; }
+    internal ConfigurationAssemblySource? ConfigurationAssemblySource { get; }
```
sungam3r, line 61: Are they mutually exclusive?

sungam3r (approved):

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationContext.cs
0xced: Yes. They influence how the assemblies that contains sinks and other types are discovered. At least one must be specified. There are three matching `ConfigurationContext` constructors and then the configuration reader is created depending on how the context was created:

```csharp
var configurationReader = context switch
{
    { ConfigurationAssemblySource: {} } => GetConfigurationReader(configuration, context, context.ConfigurationAssemblySource.Value),
    { Assemblies: {} } => GetConfigurationReader(configuration, context, context.Assemblies),
    _ => GetConfigurationReader(configuration, context ?? new ConfigurationContext(), context?.DependencyContext),
};
```

README.md
```diff
@@ -65,8 +65,9 @@ Root section name can be changed:
 ```
 
 ```csharp
+var context = new ConfigurationContext { SectionName = "CustomSection" };
```
nblumhardt, line 68: `new ConfigurationReaderOptions()`? The "context" suffix gets a bit overloaded, especially when several of the members are just regular old options.

README.md
```diff
@@ -282,6 +285,8 @@ Some Serilog packages require a reference to a logger configuration object. The
 
 When the configuration specifies a discrete value for a parameter (such as a string literal), the package will attempt to convert that value to the target method's declared CLR type of the parameter. Additional explicit handling is provided for parsing strings to `Uri`, `TimeSpan`, `enum`, arrays and custom collections.
 
+Since version 3.5.0, conversion will use the invariant culture (`CultureInfo.InvariantCulture`) as long as the `ReadFrom.Configuration(IConfiguration configuration, ConfigurationContext context)` method is used. Obsolete methods use the current culture to preserve backward compatibility.
```
nblumhardt, line 288: A lot of people will still want to write `ReadFrom.Configuration(config)` and skip the redundant extra argument. A breaking change might be fair, here, given that the source for most settings coming from configuration will be JSON, where anything other than the invariant culture is probably just harboring bugs.

What do you think?

src/Serilog.Settings.Configuration/ConfigurationLoggerConfigurationExtensions.cs
```diff
@@ -118,6 +114,7 @@ public static class ConfigurationLoggerConfigurationExtensions
     /// <param name="sectionName">A section name for section which contains a Serilog section.</param>
     /// <param name="configurationAssemblySource">Defines how the package identifies assemblies to scan for sinks and other types.</param>
     /// <returns>An object allowing configuration to continue.</returns>
+    [Obsolete("Use ReadFrom.Configuration(IConfiguration configuration, ConfigurationContext context) instead.")]
```
nblumhardt, line 117: Won't need to obsolete this if we change the default behavior.

nblumhardt (commented):
Looks great; had some thoughts on naming and defaults

README.md
0xced: I must admit that I anticipated on my next pull request which will address #206. I plan to add a new property to get access to the log level switches on that new object:

```csharp
public IReadOnlyDictionary<string, LoggingLevelSwitch> LogLevelSwitches { get; }
```

I already have a [LogLevelSwitches](https://github.com/0xced/serilog-settings-configuration/tree/LogLevelSwitches) branch on top of this `FormatProvider` branch if you want to have a look.

At that point, that `ConfigurationReaderOptions` would become «bidirectional», i.e. properties to tweak the reader behavior + property to access things (log level switches for now) created after the configuration is read. So naming it _context_ might be more appropriate I think. Maybe `ConfigurationReaderContext` context would be a better choice?

README.md
0xced: Actually, the newly introduced method already supports calling `ReadFrom.Configuration(config)` without an extra argument since the `context` parameter defaults to `null`:

```csharp
public static LoggerConfiguration Configuration(
    this LoggerSettingsConfiguration settingConfiguration,
    IConfiguration configuration,
    ConfigurationContext context = null)
```

By the way, I had to remove the default `= null` in the overload taking a `DependencyContext`. That's what I tried to explain in my original message but was maybe not very clear.

```csharp
public static LoggerConfiguration Configuration(
    this LoggerSettingsConfiguration settingConfiguration,
    IConfiguration configuration,
    DependencyContext dependencyContext)
```

So people just using `ReadFrom.Configuration(config)` will opt in the new behaviour (invariant culture) by updating to `Serilog.Settings.Configuration` 3.5.0 and recompiling their app. The only downside is that they will probably not even realise they opt in the new behaviour.

And people using `ReadFrom.Configuration(config, … extra argument …)` will get an obsolete warning and will be forced to have a look at the new overload and think about how they want to create the new `ConfigurationContext` object.

nblumhardt: Ah, thanks - I missed that. I'm trying to get a few minutes spare again to catch up with this, I'll try to loop back later today :-)

README.md
nblumhardt: Thanks for the reply! I think even in the "bidirectional" case, it would be more common to see callbacks on the options type, e.g.:

```csharp
new ConfigurationReaderOptions
{
    OnAddedLevelSwitch = levelSwitch => _mySwitches.Add(levelSwitch)
}
```

This would save us from needing to worry about the lifetime and thread-safety of other things attached to the options type, or having to explain that the type is bidirectional (and flag which members are "in" vs "out").

(Having named a few things `*Context` over the years, I've just about always regretted it one way or the other :-) ... for example, I'd love to go back and rename `LogContext` to the much more descriptive `LoggingScope`, or `ForContext()` to `WithTag()` or just about any other name ... context is used in so many ways that it's hard to guess its precise intention when you come across one.)

What do you think?

0xced: Very good point! I went with `ConfigurationReaderOptions` as you suggested and force-pushed fbc5e72b06787fd7639ab9c3179c856db34ec030 on top of the `dev` branch. I'll adapt my [LogLevelSwitches](https://github.com/0xced/serilog-settings-configuration/tree/LogLevelSwitches) branch with your `OnAddedLevelSwitch` callback proposal and I'll open a new pull request to address #206 once this one is merged.

I also adapted the README to mention _since version 4.0.0_ instead of _since version 3.5.0_ as I initially did before the rebase.

Side note: how would you name the [HttpContext ](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.httpcontext) class? 😂

nblumhardt:
Awesome PR 👍

Issue #331, opened 2022-10-11
Microsoft.Extensions.DependencyModel@3.0.0 › Newtonsoft.Json@9.0.1 security issue


joel-janerup-visma:
Affected versions of Newtonsoft.Json@9.0.1 are vulnerable to Insecure Defaults due to improper handling of StackOverFlow exception (SOE) whenever nested expressions are being processed. Exploiting this vulnerability results in Denial Of Service (DoS), and it is exploitable when an attacker sends 5 requests that cause SOE in time frame of 5 minutes.

Note: This vulnerability is only applicable to systems deployed on IIS (Internet Information Services) web-server

Fixed in https://www.nuget.org/packages/Microsoft.Extensions.DependencyModel/5.0.0 and new versions

nblumhardt:
Seems like pushing forward the dependency version here is reasonable - not sure what downstream compat problems this might cause but we've seen some friction caused by targeting 3.0.0 so we might as well bump it up to 6.0.0.

The security issue here won't be reachable via this package (it doesn't itself parse web requests), and code targeting up-to-date web frameworks will either pull in a newer Newtonsoft.JSON or not use it as all, so I think the impact of this is likely close to zero, though it's nice to keep everything fresh 👍 

PR welcome, unless any @serilog/maintainers can spot a reason to stay with the current version?

hypdeb:
One more reason to bump `Microsoft.Extensions.DependencyModel` is that it has started to cause conflicts with version `7.0.0`.

hypdeb:
I'm unable to push a branch and create a merge request. Here is a modified `.csproj` file which seems to do the job, at least according to local build and unit tests.
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <Description>Microsoft.Extensions.Configuration (appsettings.json) support for Serilog.</Description>
    <VersionPrefix>3.5.0</VersionPrefix>
    <LangVersion>latest</LangVersion>
    <Authors>Serilog Contributors</Authors>
    <TargetFrameworks>netstandard2.0;net451;net461</TargetFrameworks>
    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
    <GenerateDocumentationFile>true</GenerateDocumentationFile>
    <AssemblyName>Serilog.Settings.Configuration</AssemblyName>
    <AssemblyOriginatorKeyFile>../../assets/Serilog.snk</AssemblyOriginatorKeyFile>
[...]

PR #351, opened 2023-02-17
Drop support for .NET Framework 4.5.1

0xced:
Support for .NET Framework 4, 4.5, and 4.5.1 [ended on January 12, 2016](https://learn.microsoft.com/en-GB/lifecycle/faq/dotnet-framework), that was 7 years ago!

Also bump Microsoft.Extensions.Configuration.Binder and Microsoft.Extensions.DependencyModel to version 6.0.0 since version 2.0.0 and 3.0.0 respectively have been deprecated as part of the [.NET Package Deprecation effort](https://github.com/dotnet/announcements/issues/217).

Finally, this change allows to remove all conditional compilation.

Note that removing a target framework is a breaking change so it will require a major version bump, i.e. 4.0.0.

### DIFF
--- a/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
+++ b/src/Serilog.Settings.Configuration/Serilog.Settings.Configuration.csproj
@@ -4,7 +4,7 @@
     <Description>Microsoft.Extensions.Configuration (appsettings.json) support for Serilog.</Description>
     <VersionPrefix>3.5.0</VersionPrefix>
     <Authors>Serilog Contributors</Authors>
-    <TargetFrameworks>netstandard2.0;net451;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AssemblyName>Serilog.Settings.Configuration</AssemblyName>
     <AssemblyOriginatorKeyFile>../../assets/Serilog.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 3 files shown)


HexHunter97:
Serilog itself has also dropped support for anything below .Net Framework 4.6.1 and .Net Standard 2.0 some time ago, so there is no expectation from that angle to keep supporting 4.5.1 either.

glen-84:
4.6.1 is also EOL, why support it?

andrewboudreau:
Will this go into a 3.5 pre-release or is it too late for that?

nblumhardt:
We should look at releasing this as 4.0.0 in the very near future 👍

andrewboudreau:
I actually built this branch out and getting the same warning with 6.0.0 that I was with 3.0.0. I had to go to 7.  I assume there is some nugget switch for >= on DependencyModel and Configuration Binder

glen-84:
@0xced Any reason not to update `Microsoft.Extensions.*` to 7.0.0?

glen-84:
According to [this](https://github.com/jbogard/MediatR/pull/852#issue-1592837281), the EOL of these packages might be bound to the corresponding framework, so perhaps 6.0.0 is best.

Having said that, we still need a solution for https://github.com/serilog/serilog-settings-configuration/issues/331#issuecomment-1309860435.

Does that lead us back to #339?

andrewboudreau:
@glen-84 end of the day the issues ends up being build images available on appveyor for net7?

andrewboudreau:
Additionally, I'm running the dependency with 3.0 on other 7.0 by ignoring the warning. Things seem to to be fine. So, 6.0 probably fine?

glen-84:
Ignoring the warning is not a solution.

glen-84:
> We should look at releasing this as 4.0.0 in the very near future 👍

@nblumhardt How near is very near? 🙂

PR #354, opened 2023-03-10
Use raw string literals

sungam3r:
In general for better readability.


0xced:
Should we add a `global.json` file to ensure that the .NET 7 SDK is used? It is required for raw string literals which are a C# 11 feature.

SimonCropp:
@0xced good idea https://github.com/serilog/serilog-settings-configuration/pull/355

PR #362, opened 2023-03-13
Allow configuration from contents of Serilog section

### LINKED ISSUE #289: Read from IConfiguration without SectionKey
I'm using Serilog.Settings.Configuration to configure Serilog, and I came across an issue.
If I have an IConfiguration that consists of the Serilog-section, how can I configure Serilog with it?
`ConfigurationLoggerConfigurationExtensions.ConfigurationSection` only takes an IConfigurationSection, whilst I have an IConfiguration. And even if I did have a section - it's marked as obsolete.
`ConfigurationLoggerConfigurationExtensions.Configuration` takes an IConfiguration, but it requires an IConfiguration that is one step back, so it can use the sectionKey to get into it.

Now, I understand that you can't just add an overload to the abovementioned method, because an overload without the sectionkey already defaults to the key being "Serilog". 
Or is there perhaps some other way that I've missed? Apart from "just provide an IConfiguration that isn't already zoomed in to the Serilog section" :)

sungam3r:
fixes #289


skomis-mm (approved):
great 👍

PR #363, opened 2023-03-13
Allow internal types and methods

### LINKED ISSUE #229: Add support for 'internal' types in custom method discovery?
As part of https://github.com/serilog/serilog-sinks-map/issues/22, I ended up creating a custom extension method for my particular case, then configuring my appsettings to use it, adding the `Using` node in the json pointing to my assembly.

However, it only seems to work when the method is exposed in a `public` class. Trying to change it to `internal` and adding a `InternalsVisibleTo` attribute in my project targeting `Serilog.Settings.Configuration` didn't work at all: my method was just ignored.

I'd prefer keeping my method `internal` as I don't want anyone else from outside calling it (it is specific to this particular project).

Can this scenario be supported?

sungam3r:
fixes #229


test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
```diff
@@ -119,6 +119,87 @@ public void ConfigurationAssembliesFromDllScanning()
...
+                }
+            }
+            """;
+
+        var builder = new ConfigurationBuilder().AddJsonString(json);
+        var config = builder.Build();
+        var log = new LoggerConfiguration()
+            .ReadFrom.Configuration(
+                configuration: config,
+                readerOptions: new ConfigurationReaderOptions(ConfigurationAssemblySource.AlwaysScanDllFiles) { AllowInternalMethods = true })
```
0xced, line 139: I was confused about the DLL scanning part. I'd just use `new ConfigurationReaderOptions { AllowInternal…` and remove `FromDllScanning` from the test name for the 3 tests that were added.

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
0xced, line 1: Maybe also add tests that ensure that internal methods are **not** found when AllowInternalTypes/AllowInternalMethods are `false`.

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReaderOptions.cs
```diff
@@ -57,6 +57,16 @@ public ConfigurationReaderOptions() : this(dependencyContext: null)
     /// </summary>
     public IFormatProvider FormatProvider { get; init; } = CultureInfo.InvariantCulture;
 
+    /// <summary>
+    /// Allows to use internal types for extension methods for sink configuration. Defaults to <see langword="false"/>.
+    /// </summary>
+    public bool AllowInternalTypes { get; set; }
```
0xced, line 63: I'd use **init** instead of **set**, like all other properties of this class. You probably don't want to change those properties after the options instance is created.

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
sungam3r: ok

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReaderOptions.cs
sungam3r: ok

test/Serilog.Settings.Configuration.Tests/ConfigurationSettingsTests.cs
sungam3r: ok

nblumhardt:
LGTM! 👍

PR #367, opened 2023-03-16
Make all JSON strings valid in tests

0xced:
So that we can get rid of the `ToValidJson` extension method.

Now that raw string literals have been introduced (in #354) having valid _readable_ JSON in strings for tests becomes easy. There is no need to escape quotes anymore with either `\"` or `""`.

This will also make it easier for newcomers to understand the tests at first glance. I had my WTF moment when I first saw JSON with single quotes. 🤔

### DIFF
--- a/test/Serilog.Settings.Configuration.Tests/ConfigurationReaderTests.cs
+++ b/test/Serilog.Settings.Configuration.Tests/ConfigurationReaderTests.cs
@@ -1,4 +1,3 @@
-using System.Globalization;
 using System.Reflection;
 using Microsoft.Extensions.Configuration;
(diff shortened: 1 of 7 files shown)


0xced:
In the second commit I have also enabled JSON language injection (with `// language=json` comments) so that Rider can perform nice syntax highlighting. See [Language injections in C#][1] in Rider.

How nice is it?

image

[1]: https://www.jetbrains.com/help/rider/Language_Injections.html

sungam3r (approved):

sungam3r:
VS has syntax highlighting for JSON in raw string literals OOTB.

PR #375, opened 2023-04-20
Added missing closing bracket on Complex parameter value binding summary

DevAlvaroF:
Fix on the sample that wouldn't allow code to run by just copying it


nblumhardt:
Thanks 👍

PR #343, opened 2022-12-19
7.0.0 Release

nblumhardt:
* #233 - minimum level configuration improvements (@HexHunter97)
 * #301 - support for delegate type arguments (@almostchristian)
 * #326 - README improvements (@nblumhardt)
 * #333 - extension method discovery that goes easier on Dynatrace (@skomis-mm)
 * #334 - case-insensitive level support (@skomis-mm)
 * #342 - implicit `using` and file-scoped namespaces (@sungam3r)
 * #344 - clean-up (@sungam3r)
 * #351 - drop .NET 4.5.1 support (@0xced)
 * #348 - additional `IFormatProvider` support (@0xced)
 * #354, #357 - improve tests using raw string literals (@sungam3r)
 * #352 - `OnLevelSwitchCreated` callback (@0xced)
 * #355 - add `global.json` (@SimonCropp)
 * #356 - use `Directory.Build.props` (@SimonCropp)
 * #359 - update test refs (@SimonCropp)
 * #360 - remove redundant project settings (@SimonCropp)
 * #358 - build on macOS (@SimonCropp)
 * #361 - add API approval tests (@sungam3r)
 * #362 - support for reading configuration from `IConfigurationSection` contents (@sungam3r)
 * #364 - nullable reference type fixes (@0xced)
 * #367 - improve tests using raw string literals (@0xced)
 * #369 - more nullable reference type safety (@0xced)
 * #363 - allow configuration using internal types and methods (@sungam3r)
 * #368 - improve `SelfLog` output (@sungam3r)
 * #375 - fix README example (@DevAlvaroF)
 * #377 - pin to `Microsoft.Extensions.Configuration` v7 (@nblumhardt)
 * #353 - improve support for single-file apps (@0xced)
 * #379 - simplify assembly loading (@0xced)
 * #366 - reader options callback to expose `LoggingLevelSwitch` (@0xced)
 * #380, #381 - fix tests on .NET 4.8 (@0xced)
 * #378 - fix error handling in build script (@0xced)

PR #377, opened 2023-05-03
v7.0 - pin to MEC v7, including matching target frameworks

nblumhardt:
See also: https://github.com/serilog/serilog-extensions-logging/pull/222, https://github.com/serilog/serilog-extensions-hosting/pull/71, https://github.com/serilog/serilog-aspnetcore/pull/325

Unfortunately we've had some maintainability problems given the wide spread of target frameworks and dependencies across the Serilog sub-projects.

My aim's to kick through some RTM versions of all of these packages, pinned to their latest Microsoft.Extensions.* counterparts, ASAP.

I know we have https://github.com/serilog/serilog-settings-configuration/pull/343 open here (I've also held up the works there), my proposal is that we "rip the Band-Aid off" and merge this, then merge the release PR as v7.

Note that targeting v7 of the Microsoft.Extensions.* packages doesn't mean dropping earlier TFMs - the majority of the ME* packages continue supporting net462+, netstandard2.0+, and net6.0+.

On the release of .NET 8, the plan would then be to sim-ship Serilog.Extensions.* and Serilog.Settings.Configuration v8.0 versions that depend on the v8.0 dependencies (and so on, for the foreseeable future).

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -129,7 +129,7 @@ class ConfigurationReader : IConfigurationReader
             }
             else
             {
-                var initialLevel = ParseLogEventLevel(switchInitialLevel);
+                var initialLevel = ParseLogEventLevel(switchInitialLevel!);
                 newSwitch = new LoggingLevelSwitch(initialLevel);
             }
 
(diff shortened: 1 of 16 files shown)


Numpsy:
> On the release of .NET 8

Would the versions built with the v7 libs roll forward to work with the v8 libs (say, if you wanted to test with ASP.Net Core 8 before it's released), or does 'pinned' actually mean '7 only' ?

nblumhardt:
@Numpsy thanks for checking this out! Roll-forward would work exactly as it does today. All I mean by "pinned" is that v7 of this library will have a dependency on v7 of Microsoft.Extensions.Configuration, and v8 of this library will build against v8 of M.E.C.

This probably sounds like a bigger change than it is - the main benefit is that the package's dependency graph is predictable and consistent, irrespective of the target framework and any other factors.

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -293,7 +293,8 @@ void ApplyEnrichment(LoggerConfiguration loggerConfiguration)
         {
             foreach (var enrichPropertyDirective in propertiesDirective.GetChildren())
             {
-                loggerConfiguration.Enrich.WithProperty(enrichPropertyDirective.Key, enrichPropertyDirective.Value);
+                // Null is an acceptable value here; annotations on Serilog need updating.
```
sungam3r, line 296: https://github.com/serilog/serilog/pull/1907

PR #379, opened 2023-05-05
Simplify loading assemblies

0xced:
Use a HashSet instead of a Dictionary keyed by the assembly full name. If an assembly is loaded twice with the same AssemblyName it will be the same instance so a HashSet (without even a custom IEqualityComparer<Assembly>) is the perfect solution.

Also, Assembly.Load can throw many exceptions but won't return null.

This was initially introduced in 74402f836148f746f05a06f87cd8b15fee387155 but merged incorrectly in ac1b8c863307ceb598fbd6457ce9714ff2d1bde3.


nblumhardt:
Ah, sorry about that! 😄

PR #381, opened 2023-05-08
Fix tests on .NET Framework 4.8 (really)

0xced:
Was getting this exception because of assembly version mismatch:
> System.IO.FileLoadException: Could not load file or assembly 'System.Runtime.CompilerServices.Unsafe, Version=4.0.4.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)

Automatically generating binding redirects fixes the issue.


skomis-mm:
Thanks 👍

Issue #214, opened 2020-03-29
Allow an array-style config format for Serilog:MinimumLevel:Override


jez9999:
Starting afresh because the other issue was closed.

At the moment, the way you specify Serilog's minimum level overrides in configuration is in the following format:

	"MinimumLevel": {
		"Default": "Information",
		"Override": {
			"Microsoft.AspNetCore.Mvc.Infrastructure": "Debug",
			"Microsoft": "Warning",
			"Microsoft.Hosting.Lifetime": "Information",
			"Microsoft.AspNetCore.Authentication": "Verbose",
		}
	}

The source context (or 'namespace') is specified in the setting name, rather than a value.  The trouble with this is that not all configuration providers lend themselves to having long, complex configuration key names.  For example, I've tried to configure the equivalent to this on an Azure web app (using the double-underscore namespace separator as a replacement for the colon), and the best I could get to was:

	"Serilog:MinimumLevel:Override:Microsoft_AspNetCore_Mvc_Infras": "Debug"

On my local machine, which is using `appsettings.json`, this is how it (correctly) looks:

	"Serilog:MinimumLevel:Override:Microsoft.AspNetCore.Mvc.Infrastructure": "Debug"

The dots have been replaced with underscores, and more importantly there is a very restrictive length limit on config key names from these environment variables.  So I think the solution is not to try and put complex/long information like a namespace in the key name, but as a key value.  Thus, I propose that Serilog support the following syntax for minimum level overrides:

	"MinimumLevel": {
		"Default": "Information",
		"Override": [
			{
				"SourceContext": "Microsoft.AspNetCore.Mvc.Infrastructure",
				"Level": "Debug"
			},
			{
				"SourceContext": "Microsoft",
				"Level": "Warning"
			},
			{
				"SourceContext": "Microsoft.Hosting.Lifetime",
				"Level": "Information"
			},
			{
				"SourceContext": "Microsoft.AspNetCore.Authentication",
				"Level": "Verbose"
			}
		]
	}

When Override is an array, it takes a set of objects with SourceContext/Level that specify the override level for the given source context.  This results in the following kind of config structure, which works even with config providers that have restrictive key names:

	"Serilog:MinimumLevel:Override:0:SourceContext": "Microsoft.AspNetCore.Mvc.Infrastructure"
	"Serilog:MinimumLevel:Override:0:Level": "Debug"

skomis-mm:
@jez9999 

[...]

PR #385, opened 2023-06-28
Add support for overrides from environment variables

sungam3r:
One of my colleagues was faced with the problem when trying to set the value of the level override from environment variables. This PR helps to do that. @skomis-mm @nblumhardt I have a suspicion that there is another (simpler) solution and I do not believe that all this time no one has come across this problem. However, fix seems to be obvious and non-breaking affecting only cases when level overrides come from environment variables. Configuration from json files works well - it's safe to use dots there.

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -160,10 +160,19 @@ class ConfigurationReader : IConfigurationReader
             loggerConfiguration.MinimumLevel.ControlledBy(globalMinimumLevelSwitch);
         }
 
+        // Level overrides may be set via environment variables. In that case it may be problematic to use dots in
+        // variable name, see https://github.com/dotnet/AspNetCore.Docs/issues/17378, so all dots should be replaces
+        // with double underscore. The consequence is that configuration becomes "nested" and GetSection("Override").GetChildren()
+        // call returns nothing since this method returns only immediate descendant configuration sub-sections.
+        // The workaround is to use AsEnumerable(true) method to retrieve all descendants and then manually replace : with .
+        var flattenDescendants = minimumLevelDirective.GetSection("Override").AsEnumerable(true).ToList();
+
         foreach (var overrideDirective in minimumLevelDirective.GetSection("Override").GetChildren())
         {
             var overridePrefix = overrideDirective.Key;
             var overridenLevelOrSwitch = overrideDirective.Value;
+            // filter out "normal" keys
+            flattenDescendants.RemoveAll(p => p.Key == overridePrefix);
             if (Enum.TryParse(overridenLevelOrSwitch, out LogEventLevel _))
             {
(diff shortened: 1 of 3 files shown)


sungam3r:
If you prefer all new code can be moved under flag like `public bool ConfigurationReaderOptions.AllowNestedLevelOverrides { get; init; }`.

skomis-mm:
@sungam3r there is [OOTB](https://github.com/dotnet/runtime/issues/35989) solution is tracking on the MEC side.
Would the official [workaround](https://github.com/dotnet/runtime/issues/87130#issuecomment-1583859511) or custom serilog configuration [extension](https://github.com/serilog/serilog-settings-configuration/issues/214#issuecomment-605684024) be sufficient for now? Since we don't know what MEC fix exactly will look like (three underscores or something else)

sungam3r:
Thank you very much for links. I was sure that this problem should already be described somewhere. Both workarounds should work, yes. The problem is that both require changes in code and configuration. I mean in format of level overrides - either using `___` as indication of special case or swithing to `ApplyCustomOverrides` and splitting key/value pair into different properties. For existing infrastructures deployed in kubernetes with hundreds of applications using some shared configuration coming from key vault or something else it will be a real nightmare to change format. Of course the best option for me is fix from MS on their side but as you see
изображение
and 
изображение

I would really like to see working and non-intrusive solution for any target platform (personally I need net5/net7), not waiting for NET 8 (or 9). As you can see Serilog can deal with it with rather trivial changes. PR fixes such env usages and effectively is no-op for other cases (json config files generally). The only thing I may be afraid of is backward compatibility for unforseen cases so I suggested to introduce option to provide ability to quickly disable new behaviour if someone need it. New design of `ConfigurationReaderOptions` is perfect for this.

skomis-mm:
@sungam3r I have no strong opinion on this except general desire fix non-serilog issues outside serilog if possible.
Let's give this some time, may get a few other opinions on this

0xced:
I don't think it's a good idea to handle this at the Serilog level. This should definitely be handled upstream.

Also it seems there's a misunderstanding in your code/test. If I add your test (`TestMinimumLevelOverridesForChildContext_With_Environment_Variables`) and change `"My_Serilog__MinimumLevel__Override__System__Threading"` to `"My_Serilog__MinimumLevel__Override__System.Threading"` then the test passes without any changes in the code. It seems you have mixed configuration hierarchy which is delimited with `:` (or `__` in environment variables) and log level names which are typically delimited with the `.` character.

Actually, the problem seems very specific to Azure (not the environment variables). See [Werner Mairl's post](https://github.com/dotnet/runtime/issues/35989#issuecomment-945577529) who reports that the `.` character is somehow replaced with a `_` character.

I was not able to reproduce this in the unit test. Calling `builder.Build().GetDebugView()` yields the following (correct) debug view:

```
Serilog:
  MinimumLevel:
    Default=Warning (EnvironmentVariablesConfigurationProvider Prefix: 'My_')
    Override:
      System=Warning (EnvironmentVariablesConfigurationProvider Prefix: 'My_')
      System.Threading=Debug (EnvironmentVariablesConfigurationProvider Prefix: 'My_')
```

PR #390, opened 2023-07-30
Modify Serilog:Using section check to be less restrictive

### LINKED ISSUE #389: `Serilog:Using` Check Seems Unnecessarily Restrictive
For an AOT-published application on .NET 8, I have the following configuration (in the Microsoft.Extensions.Configuraation sense):

```json
{
  "Serilog": {
    "MinimumLevel": "Debug"
  }
}
```

…and the rest of the Serilog configuration is done in C# code. I would like be able to change this log level at runtime. However, Serilog.Settings.Configuration fails at runtime when the only assembly to search is "Serilog". Because of how that is done, the following expected workarounds do not solve the problem:

```json
{
  "Serilog": {
    "Using": [],
    "MinimumLevel": "Debug"
  }
}
```

```json
{
  "Serilog": {
    "Using": null,
    "MinimumLevel": "Debug"
  }
}
```

```json
{
  "Serilog": {
    "Using": ["Serilog"],
    "MinimumLevel": "Debug"
  }
}
```

All of which fail at runtime with the same error message: "No Serilog:Using configuration section is defined and no Serilog assemblies were found. This is most likely because the application is published as single-file." and so on.

In order to continue testing AOT publication, I have to configure like this:

```json
{
  "Serilog": {
    "Using": ["<name of entry assembly>"],
    "MinimumLevel": "Debug"
  }
}
```

…but this is undesirable since we'd have to put the name of the entry assembly in the configuration for each application. Is there or can there be a way to give Serilog.Settings.Configuration a positive signal that the single assembly it found is OK?

sungam3r:
fixes #389

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -383,7 +383,7 @@ class ConfigurationReader : IConfigurationReader
             assemblies.Add(assumed);
         }
 
-        if (assemblies.Count == 1)
+        if (assemblies.Count == 1 && !UsingSectionExists())
         {
             var message = $"""
                 No {usingSection.Path} configuration section is defined and no Serilog assemblies were found.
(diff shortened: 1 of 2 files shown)


0xced:
Based on discussion in #389 I think we should consider #391 instead of this pull request.

PR #396, opened 2023-08-28
Document destructuring configuration

satrapu:


README.md
```diff
@@ -313,6 +313,80 @@ Some Serilog packages require a reference to a logger configuration object. The
...
+{
+  public class MyDto
+  {
+    public int Id {get; set;}
+    public int Name {get; set;}
+  }
+
+  public class FirstDestructuringPolicy : IDestructuringPolicy
+  {
+    public bool TryDestructure(object value, ILogEventPropertyValueFactory propertyValueFactory, out LogEventPropertyValue result)
```
nblumhardt, line 356: Some of these lines will cause scrollbars to show up in the README - would it be possible to manually wrap them?

README.md
```diff
@@ -313,6 +313,80 @@ Some Serilog packages require a reference to a logger configuration object. The
...
+      "policy": "policy": "MyThirdNamespace.ThirdDestructuringPolicy, MyThirdAssembly"
+    }
+  },
+],
+```
+
+This is how the first destructuring policy would look like:
+
+```csharp
+namespace MyFirstNamespace
```
nblumhardt, line 346: Using a namespace statement will reduce nesting here.

README.md
```diff
@@ -313,6 +313,80 @@ Some Serilog packages require a reference to a logger configuration object. The
...
+{
+  public class MyDto
```
nblumhardt, line 348: Rest of the C# samples use four space indent.

README.md
```diff
@@ -313,6 +313,80 @@ Some Serilog packages require a reference to a logger configuration object. The
...
+  {
```
nblumhardt, line 349: Not sure how much of the C# world is on board with records yet, but:

```csharp
public record MyDto(int Id, int Name);
```

would cut a few lines?

README.md
```diff
@@ -313,6 +313,80 @@ Some Serilog packages require a reference to a logger configuration object. The
...
+    public bool TryDestructure(object value, ILogEventPropertyValueFactory propertyValueFactory, out LogEventPropertyValue result)
+    {
+      result = null;
+      MyDto dto = value as MyDto;
+
+      if (dto == null)
+      {
+          return false;
+      }
```
nblumhardt, line 364: All of the above can be replaced with:

```csharp
if (value is not MyDto dto)
{
    result = null;
    return false;
}
```

README.md
nblumhardt, line 356: The signature of `TryDestructure` might need nullable annotations to compile for many people; it's:

```csharp
bool TryDestructure(object value, ILogEventPropertyValueFactory propertyValueFactory, [NotNullWhen(true)] out LogEventPropertyValue? result)
```

README.md
```diff
@@ -313,6 +313,80 @@ Some Serilog packages require a reference to a logger configuration object. The
 },
 ```
 
+### Destructuring
+
+Destructuring means extracting pieces of information from an object and create properties with values; Serilog offers the [@ destructuring operator](https://github.com/serilog/serilog/wiki/Structured-Data#preserving-object-structure). In case there is a need to customize the way log events are serialized (e.g., hide property values or replace them with something else), one can define several destructuring policies, like this:
```
nblumhardt, line 318: We've taken to calling `@` the "structure-capturing operator", might be nicee to use that here. (Why the feature is called "destructuring" is a bit of a long story 😅)

nblumhardt, line 318: Backticks for `@`?

nblumhardt (commented):
Looks good, thanks for putting this together!

Tried to do a thorough nitpicky pass over it, wrote up a few ideas/comments 👍

satrapu:
Hey @nblumhardt,

Thanks for your feedback - all suggestions have been addressed.

README.md
```diff
@@ -313,6 +313,74 @@ Some Serilog packages require a reference to a logger configuration object. The
...
+
+        return true;
+    }
+}
+```
+
+Assuming Serilog needs to destructure an argument of type **MyDto** when handling a log event:
+
+```csharp
+logger.LogInformation("About to process input: {@MyDto} ...", myDto);
```
nblumhardt, line 375: Sorry, I missed this: the Serilog `ILogger`'s method is called `Information()`, not `LogInformation()`.

I'll edit in-place after a merge to save another round-trip.

Otherwise, all looks awesome 👍  - thanks @satrapu!

satrapu: Good job spotting that error!

PR #400, opened 2023-10-23
.NET 8 RC2

nblumhardt:
Working my way through these in preparation for RTM in a couple of weeks.

I tried to keep this one as additive/nondestructive as possible, given all the quirks this project has to deal with across the supported TFMs :-)

### DIFF
--- a/sample/Sample/Sample.csproj
+++ b/sample/Sample/Sample.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net7.0;net462</TargetFrameworks>
+    <TargetFrameworks>net6.0;net7.0;net462;net8.0</TargetFrameworks>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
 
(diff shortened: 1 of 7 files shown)


0xced:
In order to fix the tests that do not pass when running on .NET 8, `<SelfContained>true</SelfContained>` must be added to the `TestApp.csproj` file. This is because of a breaking change that was introduced in the .NET 8 SDK: [Runtime-specific apps no longer self-contained](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/runtimespecific-app-default). Especially this change:
> If `PublishSelfContained` isn't explicitly set to `false`, the publish properties `PublishSingleFile` and `PublishAot` now imply `SelfContained` (if it's not specified) during `dotnet publish` only (that is, not for `dotnet build` or `dotnet restore`).

I had to read that sentence a few times to really grasp it!

sample/Sample/Sample.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net7.0;net462</TargetFrameworks>
+    <TargetFrameworks>net6.0;net7.0;net462;net8.0</TargetFrameworks>
```
0xced, line 4: nit: `net6.0;net7.0;net8.0;net462` would be easier to the eyes than `net6.0;net7.0;net462;net8.0`

test/Serilog.Settings.Configuration.Tests/Serilog.Settings.Configuration.Tests.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT'">net48</TargetFrameworks>
-    <TargetFrameworks>$(TargetFrameworks);net7.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>$(TargetFrameworks);net7.0;net6.0;net8.0</TargetFrameworks>
```
0xced, line 5: nit: The order of the target frameworks is also a bit unsettling here.

nblumhardt:
Thanks for checking this one out @0xced, I'll take a look and fix it up 👍

PR #401, opened 2023-11-14
Update to .NET 8 RTM and Serilog 3.1.1

nblumhardt:


nblumhardt:
Working my way through the 8.0.0 releases; this is pretty low in the dependency graph so pushing this through :-)

nblumhardt:
If anyone has a chance to check out the build/test failures - I'm not very familiar with how TestApp is set up/used, it'd be great to have some more eyes on it :-)

0xced:
I'll look into it.

nblumhardt:
Think I have it sorted - thanks @0xced!

nblumhardt:
_Build.ps1_ was using a direct .dll invocation to kick off tests, something in that mode prevents `[CallerFilePath]` from being compiled in correctly; running `dotnet test` against the CSPROJ is fine.

global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "8.0.100-rc.2.23502.2"
+    "version": "8.0.100"
```
0xced, line 3: We should probably restore `"rollForward": "latestFeature"` (and maybe `"allowPrerelease": false`).

nblumhardt: Sounds good, will do on the next round if you don't beat me to it :-) 👍

PR #403, opened 2023-11-15
Restore parallel tests execution

0xced:
This reduces the tests execution time in half. On my machine: 76 seconds when run in parallel vs 144 seconds when run serially.

The fix introduced in 41bc33d4bc7b03c920102ddaba195e96d152fb5f (running `dotnet test` on the csproj rather then on the built dlls) was problematic for two reasons.

1. It rebuilds `Serilog.Settings.Configuration.Tests.csproj` in Debug instead of Release and it's unnecessary since it was previously built.
2. It doesn't run tests in parallel anymore, increasing the total tests execution time.

The _right_ fix was to disable deterministic source paths for the test project with `<DeterministicSourcePaths>false</DeterministicSourcePaths>`. This way, the `[CallerFilePath]` attribute works as expected with absolute paths on the current machine instead of being substituted with `/_/` (which was causing the test failures). This looks like a breaking change that was introduced in the .NET 8 SDK.

Also, I added back allowPrerelease = false and rollForward = latestFeature in the global.json file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -201,4 +201,3 @@ FakesAssemblies/
 project.lock.json
 
 artifacts/
-/test/TestApp-*
--- a/Build.ps1
+++ b/Build.ps1
@@ -30,6 +30,9 @@ if($LASTEXITCODE -ne 0) { throw 'pack failed' }
 
 Write-Output "build: Testing"
 
-& dotnet test test\Serilog.Settings.Configuration.Tests\Serilog.Settings.Configuration.Tests.csproj
+# Dotnet test doesn't run separate TargetFrameworks in parallel: https://github.com/dotnet/sdk/issues/19147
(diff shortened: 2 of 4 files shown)


nblumhardt:
Awesome! How on earth did you find that? :tada:

0xced:
Well, I've become kind of an MSBuild expert in spite of myself over the years. The `/_/` path was a great hint towards SourceLink, so I looked at SourceLink related stuff with [MSBuild Binary and Structured Log Viewer](https://msbuildlog.com) (thanks again @KirillOsenkov 🙏) and eventually discovered this property.

Also, I was pretty motivated because I already had spend way too much figuring out how to run test in parallel across target frameworks. 😄

KirillOsenkov:
Yes, .NET 8 enabled SourceLink by default. Happy the tool is helpful!

PR #413, opened 2024-03-15
Prefer "MinimumLevel:Default" value if "MinimumLevel" value returns empty string

### LINKED ISSUE #332: Aspnetcore 7 WebApplicationFactory doesn't get default minimum configuration correctly
Checking for the minimum section returns `""` instead of null, so the logic doesn't get the value from `Default`

DavidAllardyce:
Use value from `Serilog:MinimumLevel:Default` if value from `Serilog:MinimumLevel` is an empty string.

Fixes #332

### DIFF
--- a/test/Serilog.Settings.Configuration.Tests/LoggerConfigurationExtensionsTests.cs
+++ b/test/Serilog.Settings.Configuration.Tests/LoggerConfigurationExtensionsTests.cs
@@ -142,7 +142,7 @@ public class LoggerConfigurationExtensionsTests
 
     [Fact]
     [Trait("BugFix", "https://github.com/serilog/serilog-settings-configuration/issues/332")]
-    public void ReadFromConfigurationThrowsWhenMinimumLevelDefaultSetButMinimumLevelValueIsEmptyString()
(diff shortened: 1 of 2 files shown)


test/Serilog.Settings.Configuration.Tests/LoggerConfigurationExtensionsTests.cs
```diff
@@ -139,4 +139,23 @@ public void ReadFromConfigurationSectionDoesNotThrowWhenTryingToCallConfiguratio
             .CreateLogger();
 
     }
+
+    [Fact]
+    [Trait("BugFix", "https://github.com/serilog/serilog-settings-configuration/issues/332")]
+    public void ReadFromConfigurationThrowsWhenMinimumLevelDefaultSetButMinimumLevelValueIsEmptyString()
```
sungam3r, line 145:
```suggestion
    public void ReadFromConfiguration_ShouldNot_Throw_When_MinimumLevel_Default_Set_But_MinimumLevel_Value_Is_Empty_String()
```

sungam3r (approved):

nblumhardt:
Thanks for this @DavidAllardyce, sorry about the slow turnaround. I'd guess there are other places where this kind of behavior might affect us, which is a bit worrying, but the fix for this case looks contained/low-risk so let's get it into 8.0.3 👍

PR #432, opened 2024-08-06
Make string to enum conversion case insensitive

0xced:
All other enum parsing performed in the ConfigurationReader is also case insensitive.

Fixes #431

PR #430, opened 2024-08-03
Use consistent binding logic for nested properties in object argument values

nblumhardt:
Fixes #417 

When an argument value is a complex object, instead of deferring to `Microsoft.Extensions.Configuration` we'll now recursively use the same logic to construct its properties that we do when constructing the object itself. This means that values accepted for constructor parameters and method arguments now line up with those accepted for property setters.

This should allow for some simplification of the rest of `ObjectArgumentValue` once we've shaken out any issues.

Thanks for your patience and effort narrowing this down, @MrPsi - sorry it took a while to get back around to it! I think the limitation here is the one to fix, so if/when this is merged I think it will supersede https://github.com/serilog/serilog-sinks-email/pull/131.

PR #438, opened 2024-10-07
8.0.3 Release

nblumhardt:
- #430 - use consistent binding logic for nested objects in argument values (@nblumhardt)
 - #432 - make `string` to `enum` conversion case-insensitive (@0xced)
 - #413 - prefer `"MinimumLevel:Default"` if `"MinimumLevel"` is an empty string (@DavidAllardyce)


nblumhardt:
(Thanks for the nudge, @ArieGato)

PR #467, opened 2026-02-26
Load from if path scanned

### LINKED ISSUE #466: [BUG] Assembly.Load() no longer works for unreferenced dlls in .NET 5+
According to [dotnet runtime issue 62522](https://github.com/dotnet/runtime/issues/62522), you can no longer use `Assembly.Load()` to load DLLs that aren't reference by the project in .NET 5+. Here is the relevant quote:

> .NET 5/6+ has a simpler and much more strict policy about finding assemblies to load. It does not search for files on disk if they were not part of the original project. In this case the the SampleAssembly is in the same directory, but the app will not try to load it as it wasn't built with it.
> This means that Load will not work (as the default assembly resolution will not look for additional files on disk). On the other hand LoadFrom might work (since it is given a file path). As written it's depending on current directory, which is rather fragile.
> This change between .NET Framework and .NET 5 was intentional. The goal is to make application behavior more predictable - in this case, adding new files to the directory with the app doesn't change runtime behavior (unlike in .NET Framework).

The result is that if you add an unreferenced serilog DLL to your applications directory, Serilog will detect it by reading the files off of the disk, but fail to load it at the `Assembly.Load()` call and throw an exception in .NET 5+. This is a pain to debug, because it often happens only in production (for example, when a serilog dependency gets removed) and breaks logging if you are using Serilog, so you have no insights (ask me how I know).

marner2:
This commit makes so that .NET 5 won't break if there is an extra DLL in the directory. `Assembly.Load()` has different behavior in .NET 5 vs in other runtimes.

Fixes #466

I'm open to design discussions, as this seems to be a bit of a hack.

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/Assemblies/AssemblyFinder.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/Assemblies/AssemblyFinder.cs
@@ -5,7 +5,7 @@ namespace Serilog.Settings.Configuration.Assemblies;
 
 abstract class AssemblyFinder
 {
-    public abstract IReadOnlyList<AssemblyName> FindAssembliesContainingName(string nameToFind);
+    public abstract IReadOnlyList<(AssemblyName, string?)> FindAssembliesContainingName(string nameToFind);
 
     protected static bool IsCaseInsensitiveMatch(string? text, string textToFind)
     {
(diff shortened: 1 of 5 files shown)


nblumhardt:
Thanks for proposing this.

I think to go ahead, we'd need some analysis of what this might break, e.g. by changing the versions of assemblies that get picked up vs. what would have been found via `Load` 🤔 

Attempting to load the assembly first using `Load`, and then falling back to `LoadFrom` only in the case of a failure, might be a lower-risk approach?

In general, life's been better since pushing towards explicit assembly listing via C# code, and this does also seem better aligned with where the platform's heading through AoT, so I think only really low risk changes would be considered here.

I'm also not the maintainer who's spent most time grappling with these issues, so we'd ideally have @0xced's eyes on it, too :-)

PR #474, opened 2026-04-19
Add WriteTo.FallbackChain and WriteTo.Fallible support in configuration

### LINKED ISSUE #473: Support WriteTo.FallbackChain and WriteTo.Fallible from configuration
### Summary

Serilog 4.1 introduced two new sink-configuration methods — `LoggerSinkConfiguration.FallbackChain(...)` and `LoggerSinkConfiguration.Fallible(...)` — but neither can be called from `appsettings.json` / `IConfiguration` today. Attempting to use them produces `SelfLog` output along the lines of:

> Unable to find a method called FallbackChain...

### Why it doesn't work today

Both methods are *instance* methods on `LoggerSinkConfiguration` rather than extension methods, so they aren't picked up by the extension-method scanner in `ConfigurationReader.FindConfigurationExtensionMethods`. For instance methods, this package keeps a hand-maintained list of "surrogate" methods in `SurrogateConfigurationMethods`. Only `Sink` and `Logger` are listed for `WriteTo` today.

### Proposed fix

Add surrogate signatures for the two new methods:

```csharp
static LoggerConfiguration FallbackChain(
    LoggerSinkConfiguration loggerSinkConfiguration,
    Action<LoggerSinkConfiguration> configureSink,
    Action<LoggerSinkConfiguration> configureFallback,
    Action<LoggerSinkConfiguration>[]? configureSubsequentFallbacks = null)
    => loggerSinkConfiguration.FallbackChain(
        configureSink, configureFallback, configureSubsequentFallbacks ?? []);

static LoggerConfiguration Fallible(
    LoggerSinkConfiguration loggerSinkConfiguration,
    Action<LoggerSinkConfiguration> configureSink,
    ILoggingFailureListener failureListener)
    => loggerSinkConfiguration.Fallible(configureSink, failureListener);
```

The rest of the plumbing is already in place: `Action<LoggerSinkConfiguration>` and `Action<LoggerSinkConfiguration>[]` argument binding already work (they're how `WriteTo.Async`, `WriteTo.Conditional`, and similar wrappers resolve).

### Note on `params`

`FallbackChain`'s third parameter is `params Action<LoggerSinkConfiguration>[]`. Because `params` does not set `ParameterInfo.HasDefaultValue = true`, it isn't treated as omittable by `HasImplicitValueWhenNotSpecified` in `ConfigurationReader`. The surrogate explicitly uses `= null` to make it optional. This is the same underlying mechanical issue tracked in #441; a broader fix there would let the surrogate drop the `?? []` workaround, but the surrogates can ship independently.

### Not applicable to `AuditTo`

`LoggerAuditSinkConfiguration` does not expose `FallbackChain` or `Fallible` in Serilog. Audit pipelines are intentionally fail-loudly, so silently rerouting to a fallback or swallowing failures through a listener would break the audit guarantee. Surrogates should not be added there.

ArieGato:
Closes #473.

## Summary

- Add `FallbackChain` and `Fallible` surrogates to `SurrogateConfigurationMethods` so the two `LoggerSinkConfiguration` instance methods introduced in Serilog 4.1 can be used from `appsettings.json` / `IConfiguration`.
- Extend the `Sample` project with an `AlwaysFailingSink` plus a `SampleFailureListener` so both sections in the sample `appsettings.json` exercise the fallback path and failure-listener path at runtime.
- Add three configuration-level tests covering: primary → fallback routing, subsequent fallbacks via `params` array, and `Fallible` routing failures to a listener.

## Implementation notes

Both methods are instance methods on `LoggerSinkConfiguration` rather than extension methods, so the extension-method scanner in `ConfigurationReader.FindConfigurationExtensionMethods` doesn't discover them — same as the existing `Sink` / `Logger` surrogates.

`FallbackChain`'s third parameter is `params Action<LoggerSinkConfiguration>[]`. Because `params` does not set `ParameterInfo.HasDefaultValue = true`, it isn't treated as omittable by `HasImplicitValueWhenNotSpecified`, so the surrogate uses `= null` explicitly. Same underlying mechanical issue as #441; that broader fix would let the surrogate drop the `?? []` workaround later.

`LoggerAuditSinkConfiguration` does not expose these methods in Serilog (by design — audit pipelines are meant to fail loudly), so no corresponding `AuditTo` surrogates are added.

## Test plan

- [x] `dotnet build` passes on net462 / netstandard2.0 / net8.0 / net9.0 / net10.0
- [x] Full test suite: 325 tests on net8/net10, 326 on net9, 0 failures
- [x] Coverage on the two new surrogate methods: 100% line, 100% branch (measured with `coverlet.collector` locally)
- [x] Sample runs end-to-end and the fallback/failure-listener paths are visibly exercised (`[fallback ...]` lines in console and `[SampleFailureListener]` lines in `SelfLog`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,8 @@ Some Serilog packages require a reference to a logger configuration object. The 
 },
 ```
 
+The same nested pattern applies to wrapping sinks such as `Conditional` (from _[Serilog.Expressions](https://github.com/serilog/serilog-expressions)_), `FallbackChain`, and `Fallible` — see the [sample `appsettings.json`](sample/Sample/appsettings.json) for working examples.
+
 ### Destructuring
 
 Destructuring means extracting pieces of information from an object and create properties with values; Serilog offers the `@` [structure-capturing operator](https://github.com/serilog/serilog/wiki/Structured-Data#preserving-object-structure). In case there is a need to customize the way log events are serialized (e.g., hide property values or replace them with something else), one can define several destructuring policies, like this:
--- a/sample/Sample/AlwaysFailingSink.cs
+++ b/sample/Sample/AlwaysFailingSink.cs
@@ -1,0 +1,13 @@
+using Serilog.Core;
+using Serilog.Events;
+
+namespace Sample;
+
+public class AlwaysFailingSink : ILogEventSink
+{
+    public void Emit(LogEvent logEvent)
+    {
+        throw new InvalidOperationException(
+            "AlwaysFailingSink always throws so the sample can demonstrate fallback/failure-listener behavior.");
+    }
+}
--- a/sample/Sample/SampleFailureListener.cs
+++ b/sample/Sample/SampleFailureListener.cs
@@ -1,0 +1,22 @@
+using Serilog.Core;
+using Serilog.Debugging;
+using Serilog.Events;
+
+namespace Sample;
+
+public class SampleFailureListener : ILoggingFailureListener
+{
+    public void OnLoggingFailed(
+        object sender,
+        LoggingFailureKind kind,
+        string message,
+        IReadOnlyCollection<LogEvent>? events,
+        Exception? exception)
+    {
+        var exceptionDetail = exception is null
+            ? string.Empty
+            : $" — {exception.GetType().Name}: {exception.Message}";
+        SelfLog.WriteLine(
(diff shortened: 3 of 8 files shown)


nblumhardt (approved):
Looks good, thanks @ArieGato.

PR #478, opened 2026-05-20
Support C# 13 params collections (IEnumerable<T>, List<T>)

### LINKED ISSUE #476: Extend support for C# 13 params collections (IEnumerable, List, and Span)
### Description
Following the changes in #471, the configuration binder now gracefully handles C# 13 params collections by avoiding a NullReferenceException and returning a default value. However, as discussed, it would be a worthy improvement to actively support these collections when they are not explicitly provided in the configuration.

Currently, GetImplicitValueForNotSpecifiedKey only handles traditional params T[] arrays. This issue proposes extending that support to other collection types allowed by C# 13, [you can read more about it here](https://devblogs.microsoft.com/dotnet/csharp13-calling-methods-is-easier-and-faster/#params-collections).

### Proposed Scope
1. Generic Interfaces (IEnumerable<T>, IReadOnlyList<T>, etc.)
For parameters defined as params IEnumerable<T>, the binder should return an empty array of T. Since arrays implement these interfaces, this is the most efficient way to satisfy the requirement without extra overhead.

2. Concrete Generic Collections (List<T>, HashSet<T>, etc.)
For concrete collections, the binder could use `Activator.CreateInstance(parameter.ParameterType)` to return an empty instance of the requested collection type. This handles the majority of standard collection classes that possess a parameterless constructor.

3. The ReadOnlySpan<T> Challenge
params `ReadOnlySpan<T>` and params `Span<T>` are more complex because they are ref structs. 

**Discussion Needed:** While the other collection types are straightforward to implement, we need to discuss whether we should pursue support for ReadOnlySpan<T> and, if so, how.

### Use Case
This would allow sink maintainers to use more modern, high-performance C# 13 syntax (like ReadOnlySpan) or flexible interfaces (like IEnumerable) in their extension methods while maintaining full compatibility with Serilog.Settings.Configuration.

gyurebalint:
Resolves #476 

Following up on the discussion in the issue, this PR implements support for C# 13 generic `params` collections purely for the ergonomic and aesthetic benefits, while intentionally bypassing `ReadOnlySpan` to avoid reflection complexity.

### Changes Included:
*   **Generic Interfaces:** Intercepts types like `params IEnumerable<T>` and safely satisfies the method contract by returning an empty array.
*   **Concrete Collections:** Intercepts generic classes like `params List<T>` or `HashSet<T>` and uses `Activator.CreateInstance` to generate an empty collection.
*   **Legacy-Safe `ref struct` Evasion:** Added a framework-agnostic check using `IsByRefLikeAttribute` to catch types like `ReadOnlySpan<T>`. This ensures we don't attempt to pass a `ref struct` into the `MethodInfo.Invoke` Reflection pipeline, allowing it to safely fall back to `null` across all supported target frameworks.
*   **Tests:** Updated and added tests to verify the behavior for `IEnumerable`, `List`, and `ReadOnlySpan`.

### DIFF
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -467,8 +467,7 @@ class ConfigurationReader : IConfigurationReader
                 {
                     // Activator.CreateInstance is unlikely to succeed for collections lacking a parameterless constructor,
                     // or those relying on the [CollectionBuilder] attribute for initialization. 
-                    SelfLog.WriteLine(
-                        $"Unable to create an implicit instance of the params collection type '{paramType}' for parameter '{parameter.Name}' on method '{methodToInvoke.Name}'.\n{ex}");
+                    SelfLog.WriteLine($"Unable to create an implicit instance of the params collection type '{paramType}' for parameter '{parameter.Name}' on method '{methodToInvoke.Name}'.\n{ex}");
                 }
             }
         }
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -463,7 +463,13 @@ class ConfigurationReader : IConfigurationReader
                 {
                     return Activator.CreateInstance(paramType);
                 }
-                catch { }
+                catch(Exception ex)
+                {
+                    // Activator.CreateInstance is unlikely to succeed for collections lacking a parameterless constructor,
+                    // or those relying on the [CollectionBuilder] attribute for initialization. 
+                    SelfLog.WriteLine(
+                        $"Unable to create an implicit instance of the params collection type '{paramType}' for parameter '{parameter.Name}' on method '{methodToInvoke.Name}'. {ex}");
+                }
             }
         }
 
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -467,7 +467,7 @@ class ConfigurationReader : IConfigurationReader
                 {
                     // Activator.CreateInstance is unlikely to succeed for collections lacking a parameterless constructor,
                     // or those relying on the [CollectionBuilder] attribute for initialization. 
-                    SelfLog.WriteLine($"Unable to create an implicit instance of the params collection type '{paramType}' for parameter '{parameter.Name}' on method '{methodToInvoke.Name}'.\n{ex}");
+                    SelfLog.WriteLine($"Unable to create an implicit instance of the params collection type `{paramType}` for parameter `{parameter.Name}` on method `{methodToInvoke.Name}`: {ex}");
                 }
             }
         }
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -468,7 +468,7 @@ class ConfigurationReader : IConfigurationReader
                     // Activator.CreateInstance is unlikely to succeed for collections lacking a parameterless constructor,
                     // or those relying on the [CollectionBuilder] attribute for initialization. 
                     SelfLog.WriteLine(
-                        $"Unable to create an implicit instance of the params collection type '{paramType}' for parameter '{parameter.Name}' on method '{methodToInvoke.Name}'. {ex}");
+                        $"Unable to create an implicit instance of the params collection type '{paramType}' for parameter '{parameter.Name}' on method '{methodToInvoke.Name}'.\n{ex}");
                 }
             }
         }
--- a/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
+++ b/src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
@@ -404,8 +404,13 @@ class ConfigurationReader : IConfigurationReader
         return paramInfo.HasDefaultValue
            // parameters of type IConfiguration are implicitly populated with provided Configuration
            || paramInfo.ParameterType == typeof(IConfiguration)
-           || paramInfo.IsDefined(typeof(ParamArrayAttribute), false)
-           || paramInfo.CustomAttributes.Any(a => a.AttributeType.FullName == "System.Runtime.CompilerServices.ParamCollectionAttribute");
+           || IsParamCollection(paramInfo);
+    }
+
+    static bool IsParamCollection(ParameterInfo paramInfo)
+    {
+        return paramInfo.IsDefined(typeof(ParamArrayAttribute), false)
+            || paramInfo.CustomAttributes.Any(a => a.AttributeType.FullName == "System.Runtime.CompilerServices.ParamCollectionAttribute");
     }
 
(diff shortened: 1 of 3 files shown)


src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -431,9 +436,35 @@ static bool HasImplicitValueWhenNotSpecified(ParameterInfo paramInfo)
...
+                return Array.CreateInstance(genericTypeArg, 0);
+            }
+
+            if (paramType.IsGenericType && !paramType.IsAbstract)
+            {
+                try
+                {
+                    return Activator.CreateInstance(paramType);
+                }
+                catch { }
```
nblumhardt: In the catch block here, it might be worth pointing out:

https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/collection-expressions.md#create-methods

which describes the cases in which Activator.CreateInstance is unlikely to succeed, and also the reflection-driven API we'd need to use in order to extend capability to them.

A `Serilog.Debugging.SelfLog.WriteLine(...)` call that includes the exception message and as much information as possible about the parameter being constructed would help users figure out why these might fail.

nblumhardt (commented):
Looks good! Just one comment on diagnostics.

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
gyurebalint: Hello,
I updated the PR, here is what the log produces:
```
Unable to create an implicit instance of the params collection type 'Serilog.Settings.Configuration.Tests.BrokenCollection`1[System.String]' for parameter 'list' on method 'DummyBrokenCollection'.
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.InvalidOperationException: I am broken by design!
   at Serilog.Settings.Configuration.Tests.BrokenCollection`1..ctor() in C:\Development\serilog-settings-configuration\test\Serilog.Settings.Configuration.Tests\DummyLoggerConfigurationExtensions.cs:line 72
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean publicOnly, Boolean wrapExceptions)
   --- End of inner exception stack trace ---
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean publicOnly, Boolean wrapExceptions)
   at Serilog.Settings.Configuration.ConfigurationReader.GetImplicitValueForNotSpecifiedKey(ParameterInfo parameter, MethodInfo methodToInvoke) in C:\Development\serilog-settings-configuration\src\Serilog.Settings.Configuration\Settings\Configuration\ConfigurationReader.cs:line 464
```
Let me know if this works for you.

test/Serilog.Settings.Configuration.Tests/ConfigurationReaderTests.cs
```diff
@@ -392,4 +411,32 @@ public void ParamsSpanParameter_GracefullyReturnsDefaultValue()
...
+                AssemblyFinder.ForSource(ConfigurationAssemblySource.UseLoadedAssemblies),
+                new ConfigurationReaderOptions());
+
+            var method = typeof(BrokenLoggerConfigurationExtensions).GetMethod("DummyBrokenCollection")!;
+            var param = method.GetParameters().Last();
+
+            var result = reader.GetImplicitValueForNotSpecifiedKey(param, method);
+
+            Assert.Null(result);
+            Assert.True(logs.Count > 0, "SelfLog count was 0. The catch block was never entered!");
```
nblumhardt: The cause of the assertion failing could be unrelated to the catch block. Better to just assert the condition and drop the diagnostic. `Assert.NotEmpty(logs)` should do the job.

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
```diff
@@ -431,9 +436,40 @@ static bool HasImplicitValueWhenNotSpecified(ParameterInfo paramInfo)
...
+            {
+                try
+                {
+                    return Activator.CreateInstance(paramType);
+                }
+                catch(Exception ex)
+                {
+                    // Activator.CreateInstance is unlikely to succeed for collections lacking a parameterless constructor,
+                    // or those relying on the [CollectionBuilder] attribute for initialization. 
+                    SelfLog.WriteLine($"Unable to create an implicit instance of the params collection type '{paramType}' for parameter '{parameter.Name}' on method '{methodToInvoke.Name}'.\n{ex}");
```
nblumhardt: `SelfLog` message styles are a bit inconsistent through the codebase, but the ones in Serilog generally use format strings, backticks or no delimiters around params, and colon before trailing exception:

```
SelfLog.WriteLine($"Unable to create an implicit instance of the params collection type `{0}` for parameter `{1}` on method `{2}`: {3}", paramType, parameter.Name, methodToInvoke.Name, ex);
```

`SelfLog.WriteLine` has no overload for accepting `FormattableString` or a builder, so the `$` syntax will cause formatting to occur prematurely here.

nblumhardt (commented):
Thanks for the follow-up, just a couple of minor notes.

src/Serilog.Settings.Configuration/Settings/Configuration/ConfigurationReader.cs
gyurebalint: Thanks for the comments. <br>
The signature or Selflog.WriteLine:
`public static void WriteLine(string format, object? arg0 = null, object? arg1 = null, object? arg2 = null)
`

**A)** So I either collapse everything into that first format argument with $ or string.Format:
```
SelfLog.WriteLine($"Unable to create an implicit instance of the params collection type `{paramType}` for parameter `{parameter.Name}` on method `{methodToInvoke.Name}`: {ex}");
```
<br>

**B)** Or I extend the WriteLine function with a 4th optional `arg3 = null` parameter and do no $:
```
SelfLog.WriteLine("Unable to create an implicit instance of the params collection type `{0}` for parameter `{1}` on method `{2}`: {3}", paramType, parameter.Name, methodToInvoke.Name, ex);
```

In this serilog-settings-configuration repository option A is used everywhere. 
In the serilog repository option B is used everywhere.

How do you want to move forward? 
Let me know if I'm not seeing something here.

nblumhardt: Ah I see, thanks. Option A might be the way to go 👍

Might be worth us looking at adding explicit support to `SelfLog` for formattable strings, sometime down the track.

gyurebalint: Just out of curiosity about maintaining all these repos, when you say:
>Might be worth us looking at adding explicit support to SelfLog for formattable strings, sometime down the track.

Do you track this anywhere, so when there is a major version update you noted down this should be implemented?

Hypotheticals:
Btw this would be a performance optimization, right? We only create the string when there is an active listener.
How do you do these migrations on breaking change you just completely remove this signature and just give a new one, if they want to upgrade they would have to change their Selflogs, or do you do a 2 phase migration, first declare it obsolete, provide a newer solution and next you remove that old signature?

nblumhardt: > We only create the string when there is an active listener.

That's correct.

> How do you do these migrations on breaking change 

At this point, so many downstream sinks use those APIs that deprecation/removal isn't really a possibility. Anything we added would likely just be additive.

There probably isn't a huge pressing need to improve this, especially for this package where most code runs only once at start-up, and we only write those messages on an unlikely error path. Just mentally bookmarking it :-)

nblumhardt (approved):

nblumhardt:
Thanks! 👍

